PR #7, opened 2015-01-30
Spacer return fix

pbihler:
So that you can do something with the return value (e.g. assign a marker class to the created spacer)


suda:
Thanks for this fix!

PR #10, opened 2015-03-20
Fixed cannot use some commands problem

cakecatz:
#11


suda:
Hey @cakecatz! Thanks for this PR. Could you update it to be mergeable with current repo state?

cakecatz:
Yes, OK!

lib/toolbar-button-view.coffee
```diff
@@ -21,9 +21,17 @@ class ToolbarButtonView extends View
       @addClass iconset
 
     @on 'click', =>
+
+      @previouslyFocusedElement = $(document.activeElement);
```
jerone: Am wondering why you convert the `document.activeElement` to jQuery and later remove it from jQuery again, without using any jQuery API's...

Am also wondering why convert `previouslyFocusedElement` and `eventElement` to members instead of private variables...

cakecatz:
@jerone Now you mention it, that's right!
I'll fix it.
Thanks :smile:

PR #36, opened 2015-05-03
Fixed callback target

cakecatz:
Current `dispatch` target `document.activeElement` isn't work on some commands.
So I changed it :)

I tested this `target` with any commands.
For example, `pane:split-right`, `bookmarks:toggle-bookmark` and more.
And it worked.

### DIFF
--- a/lib/tool-bar-button-view.coffee
+++ b/lib/tool-bar-button-view.coffee
@@ -24,7 +24,7 @@ module.exports = class ToolBarButtonView extends View
         if typeof options.callback is 'string'
           atom.commands.dispatch @getPreviouslyFocusedElement(), options.callback
         else
-          options.callback(options.data)
+          options.callback(options.data, @getPreviouslyFocusedElement())
 
     @on 'mouseover', =>
       @storeFocusedElement()
--- a/lib/tool-bar-button-view.coffee
+++ b/lib/tool-bar-button-view.coffee
@@ -1,5 +1,5 @@
 {CompositeDisposable} = require 'atom'
-{View} = require 'space-pen'
+{$, View} = require 'space-pen'
 
 module.exports = class ToolBarButtonView extends View
   @content: ->
@@ -22,9 +22,12 @@ module.exports = class ToolBarButtonView extends View
     @on 'click', =>
       if not @hasClass 'disabled'
         if typeof options.callback is 'string'
-          atom.commands.dispatch atom.views.getView(atom.workspace.getActiveTextEditor()), options.callback
+          atom.commands.dispatch @getPreviouslyFocusedElement(), options.callback
         else
           options.callback(options.data)
+
+    @on 'mouseover', =>
+      @storeFocusedElement()
 
   setEnabled: (enabled) ->
     if enabled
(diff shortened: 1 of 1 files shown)


jerone:
I found this code somewhere (can't remember where) in Atom that might be better:

```coffeescript
      activeElement = document.activeElement
      # Use the workspace element view if body has focus
      if activeElement is document.body and workspaceElement = atom.views.getView(atom.workspace)
        activeElement = workspaceElement
      activeElement
```

jerone:
Ah it was here: https://github.com/atom/atom/blob/master/src/window-event-handler.coffee#L45-L51

suda:
@cakecatz could you update this PR using @jerone's example? :)

cakecatz:
@jerone Thanks :)

@suda OK!

cakecatz:
Ah, I have tried this, but activeElement become toolbar's button when clicked.
Some command target is `TextEditor`, `Workspace`. not toolbar buttons.
So I need previous activeElement.

jerone:
That seams logic. Maybe we can store the active element before button focus...

-------- Original Message --------

jerone:
Aldo I think this solution is better then the current one (because of what @cakecatz mentioned about active element always being the toolbar button), we might be thinking for a solution where developers can define their own selector, exactly as contextmenu.

cakecatz:
I try get command's `target` from `atom.commands.selectorBasedListenersByCommandName`.

    getCallbackTarget: (callbackCommand) ->
      if atom.commands.selectorBasedListenersByCommandName[callbackCommand]?
        return $(atom.commands.selectorBasedListenersByCommandName[callbackCommand][0].selector)[0]

      # inline base command
      return atom.views.getView(atom.workspace.getActiveTextEditor())

Maybe this code become to solution base.

jerone:
[`findKeyBindings`](https://github.com/atom/atom-keymap/blob/ad7566febcc835318f46fa554863b4797865f1bb/src/keymap-manager.coffee#L266) seems to be the preferred method of getting an keybinding containing a selector.

cakecatz:
@jerone Oh ,That's what I have been looking for. Thanks!

cakecatz:
I tried other solution and it's probably perfectly work :)

lib/tool-bar-button-view.coffee
```diff
@@ -34,3 +37,14 @@ module.exports = class ToolBarButtonView extends View
...
+
+  getPreviouslyFocusedElement: () ->
+    if @previouslyFocusedElement[0] and @previouslyFocusedElement[0] isnt document.body
+      @eventElement = @previouslyFocusedElement[0]
+    else
+      @eventElement = atom.views.getView(atom.workspace)
+
+  storeFocusedElement: ->
+    activeElement = $(document.activeElement)
+    if !activeElement.hasClass 'tool-bar-btn'
```
jerone: If you change this line to `if not activeElement.classList.contains 'tool-bar-btn'` you can drop jQuery. 

Was thinking of refactoring the whole package to drop jQuery. Seen some core packages that drop jQuery in favor of vanilla JavaScript.

cakecatz: > If you change this line to if not activeElement.classList.contains 'tool-bar-btn' you can drop jQuery as this isn't really necessary and only causes overhead.

Thanks! I did it :smile: 

Of course I want to switch to vanilla JS as much as possible.

jerone:
Just found out where you got the `@storeFocusedElement` function from: https://github.com/atom/atom-space-pen-views#storefocusedelement. Great find!

:+1: :shipit:

PR #41, opened 2015-05-23
Removed buttons outline on focus

nkt:
2015-05-23 10-59-33 untitled - users nkt projects pochin frontend - atom
It looks ugly.

### DIFF
--- a/styles/tool-bar.less
+++ b/styles/tool-bar.less
@@ -23,6 +23,7 @@
     }
     &:focus,
     &:hover {
+      outline: none;
       border-color: @button-border-color;
     }
   }


jerone:
What theme are you using, because I don't have this problem with any of the standard themes?

2015-05-23 10_16_53-settings - c__users_jerone_ atom - atom

Can you also list some steps to reproduce.

nkt:
`Atom Dark` + `One Dark`.
Try to click `Open...`

nkt:
Actually this is standard Chrome behaviour, which usually removes.
Here is GitHub styles for button:

jerone:
I still can't reproduce. What OS are you on? I'm on Windows. What Atom version, etc...

> Actually this is standard Chrome behaviour, which usually removes.
> Here is GitHub styles for button:

What you are showing here *isn't the default behavior*. As you can see in your screenie the blue border and box-shadow is part of GitHub style (see the linked stylesheet), which uses [Primer](http://primercss.io/). That has nothing to do with Atom, as Atom uses parts of Bootstrap.

I don't want to keep this PR from merging, but I do want to understand where that outline comes from. Setting it to `none` can conflict with themes.

nkt:
@jerone I'm using OSX. May be this is OSX specific.

jerone:
[**@nkt**](https://github.com/nkt) commented on [23 mei 2015 16:16 CEST](https://github.com/suda/tool-bar/pull/41#issuecomment-104901495 "2015-05-23T14:16:03Z - Replied by Github Comment Enhancer"):
> [@jerone](https://github.com/jerone) I'm using OSX. May be this is OSX specific.

That's probably it! I do wonder if all other buttons in Setting for example have that outline too...
Anyways it's up to @sudo to merge this. Probably a low risk of conflicting with other themes.

suda:
Yeah, this is default behaviour of Chrome + OS X. Merging.

Issue #68, opened 2015-06-06
Change tooltip position based on toolbar position


kankaristo:
Thanks for a great package, makes Atom feel a lot more like an IDE. :)

However, the tooltips for the buttons appear on top of the buttons (I have my toolbar on the left side).

[`TooltipManager::add`](https://atom.io/docs/api/v0.207.0/TooltipManager#instance-add) has an option to set the position (placement) of the tooltip.

The tooltip position should ideally be opposite of the toolbar position. So if the toolbar is at the top, the tooltip should be shown below; if the toolbar is on the left, the tooltip should be on the right.

This is what I did in `tool-bar-button-view.coffee`:
```coffee
if options.tooltip
  @prop 'title', options.tooltip
  @subscriptions.add atom.tooltips.add(
    this,
    {
      title: options.tooltip
      delay:
        show: 0
        hide: 0
      placement: 'right'  # Opposite of the toolbar's position
    }
  )
```

(I also set the delay to 0, by personal preference).

Do you think this could be made the default behavior?

jerone:
Hi @kankaristo, thanks for using this package.

I like your idea of placing the tooltips to the side when the tool-bar is vertical. On a horizontal tool-bar the tooltips are already placed correctly.

For the `delay` part, I prefer to keep this in sync with the [tabs package](https://github.com/atom/tabs/blob/18b1a46a23cd460a8f1562ddb4a3ed17c0594f50/lib/tab-view.coffee#L96-L98) or default value as with [status-bar package](https://github.com/atom/status-bar/search?utf8=%E2%9C%93&q=tooltip).

Would you like to do a PR?

kankaristo:
Pull request: https://github.com/suda/tool-bar/pull/69

kankaristo:
Fixed in b2a9e65b8825e0c598726fd767d9fae5434205ac

PR #69, opened 2015-06-06
Change tooltip placement based on toolbar position

kankaristo:
Continuing from https://github.com/suda/tool-bar/issues/68, this changes the tooltip's placement based on the toolbar's position.

The tooltip's position is opposite the toolbar, e.g. if the toolbar is on the left, the tooltip is on the right.

### DIFF
--- a/lib/tool-bar-button-view.coffee
+++ b/lib/tool-bar-button-view.coffee
@@ -12,17 +12,9 @@ module.exports = class ToolBarButtonView extends View
 
     if options.tooltip
       @prop 'title', options.tooltip
-      toolbarPosition = atom.config.get('tool-bar.position')
-      tooltipPlacement = (toolbarPosition is "Top")    and "bottom" or
-                         (toolbarPosition is "Right")  and "left"   or
-                         (toolbarPosition is "Bottom") and "top"    or
-                         (toolbarPosition is "Left")   and "right"
-      @subscriptions.add atom.tooltips.add(
-        this,
-        {
-          title: options.tooltip
-          placement: tooltipPlacement
-        }
+      @subscriptions.add atom.tooltips.add(this,
+        title: options.tooltip
+        placement: @getTooltipPlacement
       )
 
     if options.iconset
@@ -58,3 +50,11 @@ module.exports = class ToolBarButtonView extends View
   storeFocusedElement: ->
     if not document.activeElement.classList.contains 'tool-bar-btn'
       @previouslyFocusedElement = document.activeElement
+
+  getTooltipPlacement: ->
+    toolbarPosition = atom.config.get 'tool-bar.position'
+    return toolbarPosition is "Top"    and "bottom" or
+           toolbarPosition is "Right"  and "left"   or
+           toolbarPosition is "Bottom" and "top"    or
+           toolbarPosition is "Left"   and "right"
+
--- a/lib/tool-bar-button-view.coffee
+++ b/lib/tool-bar-button-view.coffee
@@ -12,7 +12,18 @@ module.exports = class ToolBarButtonView extends View
 
     if options.tooltip
       @prop 'title', options.tooltip
-      @subscriptions.add atom.tooltips.add(this, title: options.tooltip)
+      toolbarPosition = atom.config.get('tool-bar.position')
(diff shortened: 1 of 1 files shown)


kankaristo:
This sets the initial position of the tooltip correctly, but changing the toolbar's position doesn't change the tooltip's position.

I'm not sure how to handle this, do we need to loop over all buttons and their tooltips?

jerone:
[**@kankaristo**](https://github.com/kankaristo) commented on [6 jun. 2015 18:43 CEST](https://github.com/suda/tool-bar/pull/69#issuecomment-109615810 "2015-06-06T16:43:31Z - Replied by Github Comment Enhancer"):
> This sets the initial position of the tooltip correctly, but changing the toolbar's position doesn't change the tooltip's position.
> 
> I'm not sure how to handle this, do we need to loop over all buttons and their tooltips?

I think the tooltips need to be disposed and recreated when position is changed.

lib/tool-bar-button-view.coffee
```diff
@@ -12,7 +12,18 @@ module.exports = class ToolBarButtonView extends View
 
     if options.tooltip
       @prop 'title', options.tooltip
-      @subscriptions.add atom.tooltips.add(this, title: options.tooltip)
+      toolbarPosition = atom.config.get('tool-bar.position')
+      tooltipPlacement = (toolbarPosition is "Top")    and "bottom" or
```
jerone: Can you make this part into a separate function.

kankaristo:
Apparently `placement` can be a function, which is called every time the tooltip is shown, that's probably the best way to do this.

lib/tool-bar-button-view.coffee
kankaristo: Sure, I'll change `placement` to a function, so those will have to move anyway.

lib/tool-bar-button-view.coffee
```diff
@@ -12,7 +12,18 @@ module.exports = class ToolBarButtonView extends View
...
+                         (toolbarPosition is "Right")  and "left"   or
+                         (toolbarPosition is "Bottom") and "top"    or
+                         (toolbarPosition is "Left")   and "right"
+      @subscriptions.add atom.tooltips.add(
+        this,
+        {
```
jerone: The brackets here and below aren't necessary.

jerone:
[**@kankaristo**](https://github.com/kankaristo) commented on [6 jun. 2015 19:09 CEST](https://github.com/suda/tool-bar/pull/69#issuecomment-109618129 "2015-06-06T17:09:10Z - Replied by Github Comment Enhancer"):
> Apparently `placement` can be a function, which is called every time the tooltip is shown, that's probably the best way to do this.

You are correct: http://getbootstrap.com/javascript/#tooltips-options
Think that's the best solution.

lib/tool-bar-button-view.coffee
jerone: :+1:

lib/tool-bar-button-view.coffee
kankaristo: I added them for clarity (especially the `{ }`), but I can remove them. :)

jerone: I understand. we just don't use them anywhere else and like to keep the same style.
What you could do is move the `this` after `@subscriptions.add atom.tooltips.add`.

kankaristo: Should I keep the function call parentheses (like in the original)?
```coffee
@subscriptions.add atom.tooltips.add(this,
  title: options.tooltip
  placement: @setTooltipPlacement
)
```

jerone: Yes, correct me if I'm wrong, but without those the options object (`title` & `placement`), will be part of `@subscriptions.add`.

:+1:

kankaristo: Ah, that's probably right. I'm still a novice with CoffeeScript, which is why I like to have as many parentheses as possible. :P

kankaristo:
Alright, now `placement` is a function and updates automatically.

jerone:
I like it. Thanks for the PR!

kankaristo:
Thanks for merging! :+1:

Issue #49, opened 2015-05-26
Manage third-party buttons


jerone:
Third party plugins can add any buttons they want if they consume the toolbar service. Users should be able to disallow specific plugins access to the toolbar.

jerone:
Interesting: https://github.com/atom-minimap/minimap/blob/master/lib/mixins/plugin-management.coffee

cakecatz:
:heart_eyes:

cakecatz:
https://github.com/atom-minimap/minimap/blob/master/lib/mixins/plugin-management.js

jerone:
Appearently they switched from CoffeeScript to Babel. Still useful.

cakecatz:
Would we switch to Babel?

jerone:
Aldo that is a bit offtopic, I don't see any real benefit for putting the efford in switching to Babel. Everything works in CoffeeScript and Atom isn't actively converting their packages either. Unless you want to put your time into it. I'm +0

-------- Original Message --------

Issue #72, opened 2015-06-07
Add own button 


cakecatz:
Currently there is no way adding custom button on third party plugin (e.g. Label Button).
So I want method to adding own button view like this.

    button = new CustomButtonView()
    @toolBar.addCustomButton button

![screen shot 2015-06-07 at 11 28 24 pm](https://cloud.githubusercontent.com/assets/6136383/8024399/f1aa798e-0d6c-11e5-8bea-d9282b147cf0.png)

jerone:
Do mean adding a `label` to the button options or do you want a completely [different button](`https://github.com/suda/tool-bar/issues/43)?

cakecatz:
I want a completely different button :)

cakecatz:
I think this will allow add various button. (Not only button with text)
`Image Button`, `Wide Button`, and more :smile:

jerone:
Ref https://github.com/suda/tool-bar/issues/43

Issue #43, opened 2015-05-24
Custom button types (dropdown, toggle, checkbox, radio) 


nkt:
Please add support for select, like this:
![2015-05-24 22-38-39 variables less - rent - projects nkt rent](https://cloud.githubusercontent.com/assets/3505878/7789391/6eae9296-0267-11e5-8e2c-056532a54c94.png)

jerone:
What happens when you click that button? Does a dropdown menu popup? Is it a custom menu or a native select menu?

Can you supply an use-case.

Also, what app is that screenshot from?

nkt:
@jerone it's webstorm, and use-case is a build system. Select allows you choose which build system you gonna run, and other buttons is "run" and "debug".
Webstorm uses not native selects, every option has it own icon, according to build system type. I guess firstly enough just native select.

cakecatz:
:+1: I like this.

suda:
Custom button types (dropdown, checkbox, radio) are definitely something that tool-bar should support. :+1:

jerone:
[**@suda**](https://github.com/suda) commented on [26 mei 2015 12:44 CEST](https://github.com/suda/tool-bar/issues/43#issuecomment-105485205 "2015-05-26T10:44:03Z - Replied by Github Comment Enhancer"):
> Custom button types (dropdown, checkbox, radio) are definitely something that tool-bar should support. :+1:

I do agree. I was thinking for the select/dropdown to use the same styling/layout from Atom's contextmenu. Have to find out where they defined that.

cakecatz:
Maybe native select box cannot display font icons properly :cry: 
Check this FortAwesome/Font-Awesome#996

OSX 10.10. [JS Bin](http://jsbin.com/bobifahosi/1/edit?html,css,output)
![screen shot 2015-05-28 at 4 48 38 am](https://cloud.githubusercontent.com/assets/6136383/7845932/39f0426c-04f5-11e5-89a5-d1f889a8d275.png)
![screen shot 2015-05-28 at 4 48 45 am](https://cloud.githubusercontent.com/assets/6136383/7845939/41e25a0a-04f5-11e5-8cd4-4f45d3ed5c7f.png)

MaximeWillinger:
:+1: for toggle buttons associated to toggle actions !

jerone:
[...]

PR #79, opened 2015-07-07
Added custom button support

cakecatz:
This PR to allow adding variety buttons.

```
CustomButton = require './custom-button.coffee'
@toolBar.addCustomButton(new CustomButton())
```

screen shot 2015-07-07 at 5 43 06 pm

### DIFF
--- a/lib/tool-bar-manager.coffee
+++ b/lib/tool-bar-manager.coffee
@@ -9,6 +9,11 @@ module.exports = class ToolBarManager
     button.group = @group
     @toolBar.addItem button
     button
+    
+  addCustomButton: (buttonView) ->
+    buttonView.group = @group
+    @toolBar.addItem buttonView
+    buttonView
 
   addSpacer: (options) ->
     spacer = $$ -> @hr class: 'tool-bar-spacer'


jerone:
_Ref: https://github.com/suda/tool-bar/issues/72 & https://github.com/suda/tool-bar/issues/43_

So this would allow any element to be added to the toolbar which inherits from the [`View` class](https://github.com/atom/space-pen)?

I'm kind of leaning to -1 :-1: on this, as it would allow any kind of HTML in the toolbar to be added. Because any package can consume the toolbar service, they can add any HTML they like. Without the user having a way to [manage those third-party buttons](https://github.com/suda/tool-bar/issues/49), the user has no control of disabling if they don't want that. https://github.com/suda/tool-bar/issues/49 needs to be implemented first before _I_ would consider merging this PR.

I prefer to see [**defined** custom buttons](https://github.com/suda/tool-bar/issues/43) be implemented. That way we have _more_ control (not full yet) of the buttons added to the toolbar.

If I look for example to my statusbar, it's already getting cluttered with tiles I don't use or can't hide...
cluttered status-bar

/cc @suda

cakecatz:
> So this would allow any element to be added to the toolbar which inherits from the View class?

yes.

> #49 needs to be implemented first before I would consider merging this PR.

That's true.
This PR wait for #49 :smile:

> If this does get merged, we need clear rules about styling and handling of those custom buttons. A clear code example (like the one in your screenie) should also be added to the readme.

Sure.
I also like create npm module like `atom-toolbar-button` to be base of custom button.

jerone:
I'm going to close this PR for now. Let's fix #49 first or open up a PR with a [custom button](https://github.com/suda/tool-bar/issues/43).

Issue #80, opened 2015-07-08
More than one toolbar


varemenos:
Is it possible or will it be possible to add more than one toolbar?
I'd like to write a couple of tool-bar plugins that create separate tool-bars and are attached in different positions (for example one to the top and one to the right)

jerone:
It's not possible right now to add multiple tool-bars. This package only allows one tool-bar (which is placed by the user) and only allows third party plugins to add buttons (and separators).

I do like the idea, but it requires a complete rewrite of this package and it's service to support multiple tool-bars. We're always open to PRs :)

PR #81, opened 2015-07-09
added tool-bar-almighty in the list of Plugins

varemenos:


jerone:
Great to see other plugins being build!

varemenos:
It started as something different but I needed #80 so I just implemented my personal toolbar that was based on flex-tool-bar

PR #84, opened 2015-07-23
Add icon resizing triggered by window resize

mikaylathompson:
I built this for myself, but figured I'd push it back as a non-breaking change that others might find useful.

I was using atom both on my laptop and on large monitors and the most comfortable icon size was 16px in the former case and 32px in the latter.  When Automatically Resize Icons is enabled, the icon size will vary between the three options, such that it selects the largest size that fits all icons on the screen.  Also helpful if I have a few small windows open and don't want the tool bar taking up so much of the smaller space.

### DIFF
--- a/lib/tool-bar-view.coffee
+++ b/lib/tool-bar-view.coffee
@@ -61,6 +61,9 @@ module.exports = class ToolBarView extends View
     @.element.addEventListener 'scroll', @drawGutter
     window.addEventListener 'resize', @drawGutter
 
+    if atom.config.get 'tool-bar.automaticallyResizeIcons'
+      window.addEventListener 'resize', @updateAutoSize
+
   serialize: ->
 
   destroy: ->
@@ -68,6 +71,15 @@ module.exports = class ToolBarView extends View
     @detach() if @panel?
     @panel.destroy() if @panel?
     window.removeEventListener 'resize', @drawGutter
+
+  updateAutoSize: =>
+    idealSize = @height() / @items.length / 1.5
+    autoSize = 16
+    if idealSize >= 32
+      autoSize = 32
+    else if idealSize >= 24
+      autoSize = 24
+    @updateSize "#{autoSize}px"
 
   updateSize: (size) ->
     @removeClass 'tool-bar-16px tool-bar-24px tool-bar-32px'
--- a/lib/tool-bar-view.coffee
+++ b/lib/tool-bar-view.coffee
@@ -64,6 +64,14 @@ module.exports = class ToolBarView extends View
     if atom.config.get 'tool-bar.automaticallyResizeIcons'
       window.addEventListener 'resize', @updateAutoSize
 
+    atom.config.observe 'tool-bar.automaticallyResizeIcons', (nowSet) =>
+      if nowSet
+        @updateAutoSize
+        window.addEventListener 'resize', @updateAutoSize
+      else
+        @updateSize atom.config.get 'tool-bar.iconSize'
+        window.removeEventListener 'resize', @updateAutoSize
+
   serialize: ->
 
   destroy: ->
--- a/lib/tool-bar-view.coffee
+++ b/lib/tool-bar-view.coffee
@@ -71,6 +71,7 @@ module.exports = class ToolBarView extends View
     @detach() if @panel?
     @panel.destroy() if @panel?
     window.removeEventListener 'resize', @drawGutter
+    window.removeEventListener 'resize', @updateAutoSize
 
   updateAutoSize: =>
     idealSize = @height() / @items.length / 1.5
--- a/README.md
+++ b/README.md
@@ -31,6 +31,10 @@ This package provides extensible tool bar for Atom.
 * Big *(24px)*
 * Large *(32px)*
 
+### Automatic Icon Resizing
+ * Enabled *(Icons resize with the window to keep all icons visible, when possible)*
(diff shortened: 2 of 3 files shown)


lib/tool-bar-view.coffee
```diff
@@ -61,6 +61,9 @@ module.exports = class ToolBarView extends View
     @.element.addEventListener 'scroll', @drawGutter
     window.addEventListener 'resize', @drawGutter
 
+    if atom.config.get 'tool-bar.automaticallyResizeIcons'
+      window.addEventListener 'resize', @updateAutoSize
```
jerone: I miss the `removeEventListener` on destroy.

jerone: I miss the `removeEventListener` on destroy.

lib/tool-bar-view.coffee
```diff
@@ -61,6 +61,9 @@ module.exports = class ToolBarView extends View
     @.element.addEventListener 'scroll', @drawGutter
     window.addEventListener 'resize', @drawGutter
 
+    if atom.config.get 'tool-bar.automaticallyResizeIcons'
```
jerone: Can you change this to `atom.config.onDidChange` or `atom.config.observe` (which suites best) so disabling works immediately.

lib/tool-bar-view.coffee
mikaylathompson: Ah, yep, thanks.  Added now.

jerone:
Interesting idea. I don't have access to Atom for a few days. Is it possible that you can post a screencap how this works... with [LICEcap](http://www.cockos.com/licecap/) for example.

I'm also thinking why we don't include an `auto` option in the `iconSize` option...

jerone:
[**@jerone**](https://github.com/jerone) commented on [23 jul. 2015 18:52 CEST](https://github.com/suda/tool-bar/pull/84#issuecomment-124166032 "2015-07-23T16:52:19Z - Replied by Github Comment Enhancer"):
> I'm also thinking why not include an `auto` option in the `iconSize` setting...

Atom [One light UI theme ](https://github.com/atom/one-light-ui) does something similar: https://github.com/atom/one-light-ui/blob/master/lib/config.coffee#L26

lib/tool-bar-view.coffee
mikaylathompson: Interesting issues here.  In the code below, the `atom.config.observe` part works just fine (log messages print appropriately), but adding and removing event listeners doesn't seem to "stick" and whether the event is listened to or not doesn't seem to change at all.

``` coffee
    atom.config.observe 'tool-bar.automaticallyResizeIcons', ->
      if atom.config.get 'tool-bar.automaticallyResizeIcons'
        console.log "Now it's turned on."
        window.addEventListener 'resize', @updateAutoSize
        @updateAutoSize
      else
        console.log "Now it's turned off."
        window.removeEventListener 'resize', @updateAutoSize
```

It would be easy enough to keep the event listener on all the time, and instead just silence the actual updating, but definitely not quite as nice. (And is there much overhead associated with having the event listener on?)

``` coffee
  updateAutoSize: =>
    if !atom.config.get 'tool-bar.automaticallyResizeIcons'
        return
    # else continue with resizing
```

I'm going to keep playing with it and try to find some similar examples, but I haven't done much coffee/javascript, so let me know if looks very wrong here.

jerone: Have to fully read your comment, but I see a small issue with your example code that might help you further:

```
    atom.config.observe 'tool-bar.automaticallyResizeIcons', =>
      if atom.config.get 'tool-bar.automaticallyResizeIcons'
        window.addEventListener 'resize', @updateAutoSize
...
```
The `->` needs to be an `=>` as you are referencing a method from the parent with `@`.

mikaylathompson: Ohhh, yes, that was exactly it. Everything behaves as expected now. Thank you!

mikaylathompson:
And a quick screencap screeencap

jerone:
So, just had time to test this PR. Here's some feedback:
* This doesn't work for horizontal toolbars.
* The calculations need a bit refinement, most of the times it creates an overflow where buttons disappear.
* I'm a little against creating another option for this. I prefer to move this to the `Icon Size` setting with value `auto`.

I do like the logic for this feature. If you can fix above feedback I'm :+1: on merging this.

lib/tool-bar-view.coffee
jerone: :smile:

mikaylathompson:
- Yes, that's a good point; I didn't consider horizontal ones.

- I think the calculations work or not mostly depending on the number of spacers involved.  I'll look into a better way to account for that.

- I think it makes sense to move it into `Icon Size`.  I'll work on that.

Thanks for the feedback.  It'll probably be a few days before I have time to resolve this, but I plan on doing so!

suda:
Hi @mikaylathompson I would love to merge this in if you want to update it and fix issues Jeroen pointed out :)

jerone:
I'm going to close this PR for now. We're more then happy to reopen when [previous](https://github.com/suda/tool-bar/pull/84#issuecomment-124686337) mentioned feedback is handled. :bouquet:

akhilcb:
This is a great feature. Is there any chance of making this available in tool-bar package? Something like an "icon-size: auto" would help too.

PR #98, opened 2016-01-19
Remove top-level menu shortcut to fix alpha sorting

### LINKED ISSUE #97: "Tool Bar" entry in Packages menu is not sorted alphabetically, like all the other packages
If I had to guess, I'd guess maybe a leading space or something causing it to sort oddly?  In any case, it's actually really jarring to see this one out-of-order entry right at the top of that menu.

For an image see https://www.dropbox.com/s/87u83w1af1soz3y/Screen%20Shot%202016-01-19%20at%2010.19.42.png?dl=0

zertosh:
Fixes #97.

Before:
before

After:
after

### DIFF
--- a/menus/tool-bar.cson
+++ b/menus/tool-bar.cson
@@ -3,7 +3,7 @@ menu: [
     label: 'Packages'
     submenu: [
       {
-        label: '&Tool Bar'
+        label: 'Tool Bar'
         submenu: [
           { label: '&Toggle', command: 'tool-bar:toggle' }
           {


jerone:
-1 Is [maybe an Atom bug](https://github.com/suda/tool-bar/issues/97#issuecomment-172947128)...

zertosh:
Not feeling too strongly about sorting out cross platform issues.

jerone:
Thanks for your PR @zertosh. I want to keep the access keys on Windows & Linux, still looking into removing them for OSX only.

zertosh:
@jerone No worries. I understand wanting to keep the access keys.

PR #106, opened 2016-01-30
Callback modifiers. Closes #44

### LINKED ISSUE #44: Shift click for alternative action
Suggestion, make it so shift clicking an item can execute and alternate action.
For example, shift click open would open folder/add project folder, shift click save would save as, and shift click find would find in project.

jerone:
Make it possible to setup multiple modifiers for one button. With this code it's possible to execute another callback when you click the button while holding the <kbd>alt</kbd>, <kbd>ctrl</kbd> or <kbd>shift</kbd> button.

Closes https://github.com/suda/tool-bar/issues/44.

```coffeescript
@toolBar.addButton
  icon: 'octoface'
  callback: 
    '': 'application:cmd-1'  # Without modifiers is default action
    'alt': 'application:cmd-2'
    'ctrl': 'application:cmd-3'
    'shift': (data) -> console.log data  # With function callback
    'alt+shift': 'application:cmd-5'  # Multiple modifiers
    'alt+ctrl+shift': 'application:cmd-6'  # All modifiers
  data: 'Hello World'
```

When no modifier or non-defined modifier combination is pressed, the default callback (defined by the empty key) is executed.

@suda, @cakecatz, @Leonick what do you think so far?

TODO:

- [x] Specs
- [x] Documentation.
- [x] Allow callbacks too instead of only commands

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## vNext
 * Fix memory leaks and add `onDidDestroy`. Fixes #96 (by [@zertosh](https://github.com/zertosh))
 * Added support for Material Design Icons. Closes #92 (by [@JamesCoyle](https://github.com/JamesCoyle))
+* Added support for callback modifiers. Closes #44
 * Updated Font Awesome 4.5.0
 * Updated IcoMoon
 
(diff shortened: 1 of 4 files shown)


cakecatz:
Good!

jerone:
If anyone got time for a review... otherwise I'm merging this later today and we can open up PR later.

PR #109, opened 2016-02-03
update version references to 0.2.0

devoncarew:
Update some version references from `0.1.0` --> `0.2.0`. I'm not 100% sure of the rules around `providedServices` - I don't know how to say that the service supports a range of versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,12 +58,12 @@ Make sure the following properties are part of your `package.json`.
 
 ```json
 "package-dependencies": {
-  "tool-bar": "^0.1.0"
+  "tool-bar": "^0.2.0"
 },
 "consumedServices": {
   "tool-bar": {
     "versions": {
-      "^0.1.0": "consumeToolBar"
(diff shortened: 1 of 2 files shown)


jerone:
I thought about doing that, but didn't do it because there were no breaking changes between 0.1.# and 0.2.0. Then again, there is a [new feature added ](https://github.com/suda/tool-bar/pull/106) in 0.2.0. Going to look into this...

jerone:
Come to think about it, I don't think these changes aren't really required. Service versions aren't linked to the package version. And as the current service can work beautifully with the older API we don't need to release a new service version. We could keep this PR's change to the `package-dependencies`, but as this isn't implemented by Atom yet it's just redundant. Please correct me if I've missed something and I'll reopen this PR.

devoncarew:
> Service versions aren't linked to the package version

Ah, ok, I think I was conflating the two. Totally agree that it's not necessary to do this change :)

PR #111, opened 2016-02-09
Fix scrolling

simurai:
Adding the header/footer panels https://github.com/atom/atom/pull/9274 to Atom's core breaks scrolling in tool-bar. It will probably land in Atom `1.6`. Plus tool-bar overlaps any other footer packages:

screen shot 2016-02-09 at 5 09 34 pm

Most packages aren't affected, but tool-bar's selector beats [core](https://github.com/atom/atom/blob/127e4dbae404257f14e8a050c44c56964902c8ac/static/panels.less#L64-L68) in specificity.

This PR fixes it by moving the `height: 100%;` to a less specific selector. `.tool-bar.tool-bar-16px.tool-bar-vertical` -> `.tool-bar-vertical`. Once `1.6` is out, it could also be removed.

tool-bar

### DIFF
--- a/styles/tool-bar.less
+++ b/styles/tool-bar.less
@@ -77,7 +77,6 @@
     }
     &.tool-bar-vertical {
       flex-direction: column;
-      height: 100%;
       overflow-x: auto;
       &::-webkit-scrollbar {
         display: none;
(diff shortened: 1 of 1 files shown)


jerone:
So we can merge this PR today and when we support Atom 1.6.0 and above the extra lines could be removed?

simurai:
Right. Well, maybe keep it at least till `1.7.0` or longer for those that don't update Atom right away. The override in core probably is gonna stay till `2.0`. So there is no rush.

PR #116, opened 2016-02-17
:shirt: Added lint script to package.json

jerone:
For all contributors. 

Call with `npm run lint`


jerone:
When developing I suggest having the following packages installed:

* https://atom.io/packages/linter
* https://atom.io/packages/linter-markdown

PR #118, opened 2016-02-19
Increment priority by group

jerone:
This will fix random ordering buttons when no priority was specified.

The previous situation was that it copied the priority from the last button/spacer. This could be from the same group (thus package) or from an other group. Resulting in buttons with the same priority from multiple groups ordered by first come.

New situation is that it will increment the priority from the last appended button/spacer that isn't part of the current group, making sure that all button/spacer without priority are at least grouped together.

The real fix is of course just to give all buttons/spacers in your package an priority:

* [x] https://github.com/cakecatz/flex-toolbar/pull/56


cakecatz:
:+1:

PR #121, opened 2016-02-21
:sparkles: Full width header & footer tool bar. Closes #91

### LINKED ISSUE #91: Show over/under Tree View as well
Currently, the Tool Bar is shown only over/under the editor-portion of the editor.

It would be nicer if it would be shown over and under the tree-view as well (File Browser/Sidebar/whatever).

PS: I was planning to use this as an alternative to the menu-bar but without something this, it's not really good-looking. :sweat_smile:

jerone:
Closes #91

This will add a new setting to extend the tool bar full width horizontal, forcing it above or below the tree view. As the tool bar is for the whole workspace (and not just the editor pane), and all programs follow that convention, the default for this setting is `true`.

screenshot

:rotating_light: **Note:**
This PR requires https://github.com/atom/atom/pull/9274, ~~which has been merged into Atom beta, but not in stable. Let's wait for this to go public first.~~

:green_heart: ~~As expected beta channel is passing, stable is failing.~~ All green.

### DIFF
--- a/lib/tool-bar-view.coffee
+++ b/lib/tool-bar-view.coffee
@@ -2,6 +2,8 @@
 {View} = require 'space-pen'
 _ = require 'underscore-plus'
 
+supportFullWidth = typeof atom.workspace.addHeaderPanel is 'function'
+
 module.exports = class ToolBarView extends View
   @content: ->
     @div class: 'tool-bar'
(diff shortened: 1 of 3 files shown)


cakecatz:
Cool!
I'm using version `1.7.0-dev-9720541` and this works fine :smirk_cat:

jerone:
Thanks for testing this @cakecatz on Atom dev.

jerone:
I've updated the code to be backwards compatible.

Three greens :green_heart: as beta is landed.

Merging after updating docs, hopefully I've time for that tomorrow.

suda:
✨ You rock man!

PR #131, opened 2016-03-26
Plugin manager

### LINKED ISSUE #49: Manage third-party buttons

jerone:
Closes #49 

tool-bar-plugin-manager

I've been running this for a few days now and it seems to work great.

Some comments:
* Config settings are set programmatically.
* New config settings (by installing package) require an Atom restart (Settings Views cache needs to be cleared for the setting to appear).
* By default all plugins are enabled (backwards compatible).
* Disabling & enabling works instantly (see screenie).
* All plugins settings in the Settings View are grouped in a section named "Plugins".
* Groups can't have description, but I think the plugin descriptions are sufficient.
* I don't think we can remove the config setting when plugin is uninstalled ?!
* Ideas "copied" from [Minimap](https://github.com/atom-minimap/minimap/blob/master/lib/mixins/plugin-management.js) package.
* This opens support for [custom button types](https://github.com/suda/tool-bar/issues/43).

TODO:
* [X] Implement plugin manager.
* [x] Disable on start.
* [x] Hide newly added buttons.
* [ ] Make hiding & restoring of buttons private.
* [ ] Specs.
* [ ] A line mentioning this in the readme.
* [ ] Remove consoles.
* [ ] Test, test, test.
* [ ] Squash commits.

### DIFF
--- a/lib/tool-bar-manager.coffee
+++ b/lib/tool-bar-manager.coffee
@@ -3,11 +3,13 @@ ToolBarButtonView = require './tool-bar-button-view'
 
 module.exports = class ToolBarManager
   constructor: (@group, @toolBar) ->
+    @visible = true
 
   addButton: (options) ->
     button = new ToolBarButtonView options
     button.group = @group
     @toolBar.addItem button
+    @toolBar.hideItem(button) unless @visible
     button
 
   addSpacer: (options) ->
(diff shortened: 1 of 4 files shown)


jerone:
Superseded by #132

Issue #135, opened 2016-04-28
Removing space-pen


zertosh:
I was about to start removing space-pen in favor of using the DOM directly, but then I figured I'd ask first if it made sense to do a port of the current code to ES6 and then remove space-pen. If that sounds good, do you have a preference on lint rules?

jerone:
That would be so cool! It's on my own todo list for a few months now, since I saw the trend happening where other packages also switched to ES6. Combining that with removing the `space-pen` dependency would be great too, as it looks like the Atom team is deprecating that library in the near future and building [their own](https://github.com/atom/etch). For me I don't have a clear preference for ES6 linting as I've not done much. I've used `eslint-config-standard` at one of [my projects](https://github.com/jerone/atom-html2js/blob/master/package.json#L28-L29), but if you've got a better preference go for it.

**:+1: on removing `space-pen`.
:+1: on switching to ES6.**

With one remark that we don't lose any API functionality and the tests are also converted!
I can even chip in if you want.

I do think that we need @suda's permission as well!

suda:
I'm not a fan directly manipulating DOM like it's 1997 😄  But I would love if we could switch to Etch. I hope Atom team won't introduce too many breaking changes in. So in summary:

:+1: on removing `space-pen`
:+1: on switching to ES6
:+1: on using Etch

jerone:
> I'm not a fan directly manipulating DOM like it's 1997

Remember that we are working in an Chromium environment, so we can always use the latest and best API's available. jQuery isn't required here anymore.



> But I would love if we could switch to Etch

I asked on Slack about Etch and it's state. The answer was in the direction of it still being developed and API's may change. On the other side they sound confident that the direction it's headed is the right one. Just don't want to make promises. They are using it on some Atom stuff so I don't foresee it going away. It looked like they were interested in more feedback while implementing it into existing packages.

This creates the question if we should jump on it right now, or wait a little bit more until it's more mature...

jerone:
So now #136 is merged let's discuss how `space-pen` can be removed...

I'm kinda hesitated to introduce another framework, that's not even released yet and might be a bit overkill for this package. On the otherhand I might be cool if we were the first mayor package introducing Etch framework.

@BinaryMuse Do you have a opinion if the tool-bar package is a good starting point to implement Etch?

BinaryMuse:
[...]

PR #136, opened 2016-04-29
One-to-one port from CoffeeScript to ES6

zertosh:
With a few minor exceptions (like using native Array methods instead of `underscope-plus`) this is a one-to-one port from CoffeeScript to ES6. There was no refactoring or anything, just porting the code over. This is preparation for removing `space-spen`. See #135.

A note on style: I went with [`eslint-config-standard`](https://github.com/feross/eslint-config-standard) but the lack of semi-colons was a little too weird. I turned them back on in the `.eslintrc` file. I'll let @suda and @jerone bikeshed a bit on style, and I'll apply the changes here accordingly.


jerone:
Awesome work @zertosh! Skimmed through it and looking awesome. I might have time to review and test tonight (no promises).

README.md
```diff
@@ -88,72 +88,91 @@ Make sure the following properties are part of your `package.json`.
...
-      'alt': 'application:cmd-2'
-      'ctrl': 'application:cmd-3'
-      'shift': (data) -> console.log data  # With function callback
-      'alt+shift': 'application:cmd-5'  # Multiple modifiers
-      'alt+ctrl+shift': 'application:cmd-6'  # All modifiers      
+    callback: {
+      '': 'application:cmd-1',      // Without modifiers is default action
+      'alt': 'application:cmd-2',
+      'ctrl': 'application:cmd-3',  // With function callback
+      'shift'(data) {
```
jerone: Is this correctly formatted?

.coffeelintignore
```diff
@@ -1,2 +0,0 @@
-iconsets
-node_modules
```
jerone: Is there a similar option for ES6 linting for ignoring both directories?

README.md
```diff
@@ -88,72 +88,91 @@ Make sure the following properties are part of your `package.json`.
...
 
-```coffeescript
-consumeToolBar: (toolBar) ->
-  @toolBar = toolBar 'your-package-name'
+```js
+let toolBar;
 
-  # Add buttons and spacers here...
+export function consumeToolBar(toolBar) {
+  toolBar = toolBar('your-package-name');
```
jerone: Doesn't the argument `toolBar` and the global `toolBar` variable conflict with each other?

.coffeelintignore
zertosh: eslint has an equivalent `.eslintignore`. but since eslint ignores `node_modules` by default and there are no `.js` files in `iconsets`, I felt it unnecessary to add.

README.md
zertosh: yup. broke this example.

README.md
zertosh: Short methods can be computed, string literals or just identifiers. Here, you don't need the quotes for `alt`, `ctrl`, or `shift`, but I kept them so the entire object looked consistent. I can remove them though.

.coffeelintignore
jerone: Let's ignore it just in case. It still searches in there for `.js` files and we never know what happens when sets get updated or added.

README.md
jerone: Still learn new things in ES6.

.coffeelintignore
zertosh: :+1:

jerone:
I'm wondering if the linting actually runs... A list of all linted files should appear here: https://travis-ci.org/suda/tool-bar/jobs/126903567#L191 like it does with CoffeeScript https://travis-ci.org/suda/tool-bar/jobs/120148443#L180

.coffeelintignore
jerone: Never mind, it's only used when running linting locally. Linting by CI services only run `lib` & `specs` folder: https://github.com/atom/ci/blob/master/build-package.sh#L75-L86.

zertosh:
> I'm wondering if the linter actually runs...

Just added `--format tap` to the eslint command so it prints the linted files.

.coffeelintignore
zertosh: Wow... That script is news to me. So the `--format tap` flag I added to the package.json only runs locally.

jerone: Then this is probably a "bug" in the `build-package.sh` file. One way to check if linting is run is to make a temporary bug.

p.s. I'm all for squashing commits when the situation seems right (which it does in this case), but only when everything is finished and not within a review process :)

jerone:
The rewrite looks good to me.
Still wondering about ES6 linting.
Testing still required.

.eslintrc
```diff
@@ -0,0 +1,25 @@
+{
```
jerone: Can you rename this file to `.eslintrc.json` as it's current naming is deprecated per http://eslint.org/docs/user-guide/configuring#configuration-file-formats

jerone:
@zertosh Just an update. Yesterday I had time to test your code, but due to a bug in Atom (which I eventually found) I couldn't test the new code. I'll try to make some time this weekend to test it.

zertosh:
No worries @jerone!

lib/tool-bar-view.js
```diff
@@ -0,0 +1,229 @@
...
+
+  updatePosition (position) {
+    this.removeClass(
+      'tool-bar-top tool-bar-right tool-bar-bottom tool-bar-left ' +
+      'tool-bar-horizontal tool-bar-vertical'
+    );
+
+    const fullWidth = supportFullWidth && atom.config.get('tool-bar.fullWidth');
+
+    if (position === 'Top') {
```
jerone: :art: I prefer to keep the `switch` here.

jerone:
I experiencing a bug with buttons aligned at the end (negative priority). 

Steps to reproduce:

1. Install flex-tool-bar (only needed for easy testing).
2. Copy the following to the `toolbar.cson`: 
```
[
  {
    type: "button"
    icon: "file-outline"
    callback: "advanced-open-file:toggle"
    tooltip: "New File"
    iconset: "mdi"
  }
  {
    type: "button"
    icon: "content-save"
    callback: "core:save"
    tooltip: "Save"
    iconset: "mdi"
  }
  {
    type: "button"
    icon: "thumbnails"
    callback: "expose:toggle"
    tooltip: "Expose"
    iconset: "fi"
    priority: -50
  }
]
```
3. The `thumbnails` icon should be right aligned, but isn't.
4. Remove the second button, and the `thumbnails` button is right aligned again.

jerone:
> I experiencing a bug with buttons aligned at the end (negative priority).

Probably a bug in the converted code here: https://github.com/zertosh/tool-bar/blob/coffeescript-to-es6/lib/tool-bar-view.js#L78-L104

I'll try to make some failing tests later for you to implement.

jerone:
Aldo I still get different results between this PR and the master, creating specs also highlighted a bug with the priority calculation. Opening PR...

jerone:
Oke, merged #137 also containing new specs to hopefully find previously mentioned bug.
When you've got time, please rebase this PR. Ping me afterwards so I can test again :)

lib/tool-bar-view.js
zertosh: Done! I tried to mimic the short if-else syntax with ternaries. Let me know if you prefer if-else too.

zertosh:
@jerone thanks for the fix and test case! I pushed the changes rebased on top of master, but not squashed.

lib/tool-bar-view.js
jerone: This is fine now :+1:

jerone:
Hmmm, still getting same difference as [mentioned before](https://github.com/suda/tool-bar/pull/136#issuecomment-217118820). I thought the specs would have failed, but it didn't.

This happens with master:
2016-05-05 19_28_16-toolbar cson c__users_jerone_documents_github_tool-bar atom

This is the result of same `toolbar.cson` with this PR:
2016-05-05 19_28_51-toolbar cson c__users_jerone_documents_github_tool-bar atom

I don't have time to investigate, maybe you've got time. Can't merge with this difference :disappointed:

zertosh:
@jerone found the bug, sorry about that. fixed in 2d1a882.

jerone:
Lol, almost same time and almost same fix: https://github.com/zertosh/tool-bar/pull/1

If you change the `0` to `this.items.length` the tests should succeed.

zertosh:
ha! 3 minutes apart. I thought I had run the tests 😫  Anyway, I went with your syntax - though as a personal preference, I try not to reassign variables.

.eslintrc.json
```diff
@@ -0,0 +1,25 @@
...
+    "atom": false,
+    "afterEach": false,
+    "beforeEach": false,
+    "describe": false,
+    "fdescribe": false,
+    "xdescribe": false,
+    "expect": false,
+    "it": false,
+    "fit": false,
+    "xit": false,
```
jerone: Linting should actually fail on `fdescribe`, `xdescribe`, `fit` and `xit`. If one of those functions are committed that means that the specs broken.

zertosh: Those are the same globals that Atom declares, but sure. https://github.com/atom/atom/blob/e3790b8f20f96e17c2ad5cad3ed53b41e4f23377/package.json#L162-L177

jerone: Yeah I've seen them committing `fdescribe` once, but I think that's bad practice.

jerone:
Awesome work @zertosh :+1: 

Would be great if you can merge all commits, as I'm kind of done reviewing.

I've asked the others to pitch in, specifically about the lint preferences. But I'm very excited to get this in.

zertosh:
Done! Thanks for reviewing. Once this is in, I'll work on removing space-pen.

jerone:
Question, did you use a tool to convert CS to ES6? I'm asking because I created #132 which is in CS, aldo converting by hand isn't much work in this case.

zertosh:
It was by hand :( Though I found it easier to go from compiled coffeescript to ES6, than to go from CS to ES6.

zertosh:
http://coffeescript.org/ has a "Try CoffeeScript" button at that top. It gives you a live preview of the compile - I found it helpful.

jerone:
Ah yeah, I've used http://js2.coffee and Atom Preview to learn and understand CS. Only the step from JS to ES6 I'm still figuring out.

jerone:
:tada:

jerone:
:tada: Thank you so much @zertosh !

PR #140, opened 2016-05-15
[RFC] Remove space-pen

zertosh:
This is a pretty big breaking change. In [Nuclide](https://github.com/facebook/nuclide), we definitely assume that the return value of `addButton` is a space-pen view. I can imagine that this is the case for a lot of other people. Up to what point should this be mitigated? E.g.: should `ToolBarButtonView` fake having a property `0` so you can grab the underlying element like that?

There's a lot of room for improvement and performance now that space-pen is gone, but I'll do that in a follow-up PR. This is just a first-pass port.

### DIFF
--- a/lib/tool-bar-button-view.js
+++ b/lib/tool-bar-button-view.js
@@ -1,69 +1,75 @@
 'use babel';
 
 import {CompositeDisposable} from 'atom';
-import {View} from 'space-pen';
 
 let prevFocusedElm = null;
 
-export default class ToolBarButtonView extends View {
-  static content () {
-    return this.button({class: 'btn btn-default tool-bar-btn'});
-  }
+export default class ToolBarButtonView {
 
   constructor (options) {
-    super(options);
-
+    this.element = document.createElement('button');
     this.subscriptions = new CompositeDisposable();
+
     this.priority = options.priority;
     this.options = options;
 
-    if (this.priority < 0) {
-      this.addClass('tool-bar-item-align-end');
-    }
-
     if (options.tooltip) {
-      this.prop('title', options.tooltip);
+      this.element.title = options.tooltip;
       this.subscriptions.add(
-        atom.tooltips.add(this, {
+        atom.tooltips.add(this.element, {
           title: options.tooltip,
           placement: getTooltipPlacement
         })
       );
     }
 
-    if (options.iconset) {
-      this.addClass(`${options.iconset} ${options.iconset}-${options.icon}`);
-    } else {
-      this.addClass(`icon-${options.icon}`);
+    const classNames = ['btn', 'btn-default', 'tool-bar-btn'];
+    if (this.priority < 0) {
+      classNames.push('tool-bar-item-align-end');
     }
+    if (options.iconset) {
+      classNames.push(options.iconset, `${options.iconset}-${options.icon}`);
+    } else {
+      classNames.push(`icon-${options.icon}`);
+    }
+
+    this.element.classList.add(...classNames);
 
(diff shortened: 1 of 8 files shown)


jerone:
Looks like a solid replacement :+1:

jerone:
We need to make this backwards compatible for at least another major version.

One way to do that is provide another service version: 

```json
  "providedServices": {
    "tool-bar": {
      "description": "A container for buttons at workspace edge",
      "versions": {
        "0.1.0": "provideToolBarLegacy",
        "1.0.0": "provideToolBar"
      }
    }
  }
```

I've been playing to get this working by providing an jQuery object when run through the `provideToolBarLegacy` service, but I haven't had full success. This is the class to wrap around the returning button and spacer:
```es6
import jquery from 'jquery';

class space extends jquery {
  constructor(elm) {
    super(elm);
    this._elm = elm;
  }
  setEnabled(){
    this._elm.setEnabled.apply(this._elm, arguments);
  }
}
```

jerone:
Ref: https://github.com/suda/tool-bar/issues/135

zertosh:
I just added the legacy provider. The test is going to fail. I don't know why jQuery isn't respecting the event cancellation.

A few things:

* I'm not a fan of this much backwards-compat. IMHO, bumping the provider version is enough. I can't imagine that the upgrade path is that difficult. If there has to be BC, then instead of wrapping the new view or extending jquery, then the old code should be kept, and that's what is returned.

* Grim... It's big (~4mb). If you're including jQuery + Grim, might as well just include space-pen (it has both). Which goes back to doing BC by including the old code.

jerone:
We need to be backward compatible, at least for another minor version (discussable)! 

Skimming through the list of [plugins](https://atom.io/packages/search?utf8=%E2%9C%93&q=keyword%3Atool-bar) not many use the return type of the API for something useful. I'm happy to remove the legacy code as soon as those package switch to the new service provider.

I've never considered file size for dependencies an issue. Grim will be there for only the legacy code, but it lets package developer know, through Atom, about outdated API code. If there are other (smaller) solutions we can consider that.

jerone:
> Grim... It's big (~4mb).

Just spoke with the nice people of Atom and wallah: https://github.com/atom/grim/pull/13, which should minimize the file size enormous.

If we lazy load this package, I'm +1 on using it to let package developers know about the new provider service.

zertosh:
I was the one that removed `grim` from `event-kit` lol... I'll update the PR

jerone:
Seriously ... lol :+1: 

Am also looking at the one failing spec.

zertosh:
> Am also looking at the one failing spec.

I was going to just keep the legacy class since it's only temporary. That should fix it, and then I can make a few changes to the new API (like add `priority` get/set methods) without worrying about compat.

zertosh:
ugh `space-pen` still has `grim@^1.0.0`

jerone:
> ugh space-pen still has grim@^1.0.0

Which probably won't get an update anymore. Besides, I'm more then happy with the current implementation. If the last failing spec can be fixed, I'm :+1: on merging this.

jerone:
> If the last failing spec can be fixed

It seems to be an issue between `onclick` attribute & `addEventListener('click')` handler. If you change line 336 (on both spec files) to `toolBar.addEventListener('click', clickSpy);` the specs succeed.

zertosh:
I think this is good now.

jerone:
LGTM :tada:

zertosh:
derp, i was going to squash :(

jerone:
Though about it; but all commit had clear messages.

PR #143, opened 2016-05-24
Togglable buttons and new colours

MikeInnes:
This adds a simple button method which adds the `.selected` class to itself, as well as a method for marking the button as `btn-primary`, `btn-error` etc. My main use case for the latter option is for lighting up the console button as a way of notifying the user that there's new output to view.

toolbar

### DIFF
--- a/lib/tool-bar-button-view.js
+++ b/lib/tool-bar-button-view.js
@@ -50,12 +50,20 @@ export default class ToolBarButtonView {
     }
   }
 
+  isEnabled () {
+    return !this.element.classList.contains('disabled');
+  }
+
   setSelected (selected) {
     if (selected) {
       this.element.classList.add('selected');
     } else {
       this.element.classList.remove('selected');
     }
+  }
+
+  isSelected () {
+    return this.element.classList.contains('selected');
   }
 
   destroy () {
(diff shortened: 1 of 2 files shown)


jerone:
I'm kind of hesitate to merge this PR, because of three reasons:

1. This is part of https://github.com/suda/tool-bar/issues/43. Please read that topic, especially the [last post](https://github.com/suda/tool-bar/issues/43#issuecomment-199403194).
2. This all is simple to build with the return type of `addButton`.
3. Failing linting & missing specs.

MikeInnes:
Well, tests and linting are at least easy to fix. The main reason I hesitated on the specs was because it seemed like the logic is so trivial it wouldn't really add any value; you're basically testing `classList.add('foo'); classList.contains('foo');`. The fiddly / more fragile part was the CSS styling, but I don't know how to test that things look right. Of course, if you're adamant and/or have any better ideas I'm happy to do it anyway.

Although I didn't try modifying the returned value, I suspect styling would be an issue – but perhaps you'd accept just some small CSS changes to make custom styling easier?

I see the issue of the toolbar filling up, although I'm not clear on how that's related to this change. Is the assumption that if tool-bar has more functionality then more packages will make use of it? If so that seems like an odd argument. For my part I'm mainly adding styling to existing buttons, and in any case I have a toggle in the package settings to disable toolbar and statusbar integration – arguably, the onus rests on package authors to provide that, making the issue orthogonal to this PR.

PR #144, opened 2016-05-28
Update Material Design Icons to v1.6.50

unluisco:


jerone:
The test fails: https://travis-ci.org/suda/tool-bar/jobs/133624745#L229
Can you explain why it's failing.

unluisco:
The icon 'material-ui' in the old file is /f449, but in the update file is /f357

jerone:
Yes, please fix the specs if you with certainty can say that the code for the same icon is changed.

unluisco:
This PR has passed the checks

jerone:
Thanks!

PR #145, opened 2016-06-08
Added ability to add custom class names to items (both buttons and spacers) on the toolbar

rageandqq:
I noticed that there was no way to add custom classes to buttons and spacers on the toolbar. While not essential for spacers, when there are two buttons with the same icon in the toolbar (or another item that shares the same octicon elsewhere on the webpage), it becomes difficult to consistently distinguish which one is which.

This diff adds the ability to query the DOM using `document.querySelector` or the like and identify a unique element by class name (instead of depending on the class name provided by the octicon).

Tests were added to test this feature -- existing tests already handle the case when `.classNames` is missing from the `options` parameter.

Additionally, I updated the README to reflect the usage of this feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ Make sure the following properties are part of your `package.json`.
 }
 ```
 
-We recommend using [Atom-Package-Deps](https://github.com/steelbrain/package-deps) 
+We recommend using [Atom-Package-Deps](https://github.com/steelbrain/package-deps)
 in your package for installing dependency packages like this package.
 
 ### Main package file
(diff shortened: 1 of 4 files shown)


rageandqq:
Whoops, didn't realize there was a linter + rules set up!

jerone:
Haven't had time to review this, but you can use the return type of `addButton()` to do anything with the buttons. For example:
```js
toolBar.addButton({ icon: 'octoface' }).element.classList.add("custom-class-name");
```

jerone:
I'm going to close this PR for the exact reason mentioned [above](https://github.com/suda/tool-bar/pull/145#issuecomment-224514603), where the return value _can_ be used to alter the button/spacer (can is italic here, because it's an undocumented feature for now).

Issue #141, opened 2016-05-15
[WIP] Update plugin to new provider service


jerone:
I think it's nice to let plugin developers know that there's a new provider service available so we can remove the legacy code as soon as possible. Besides that, the previously documented provider semver won't support the new version we're releasing.

There are two lists of plugins:

1. Plugins that consume the service and the public API.
2. Plugins that consume the service and use the return type.

List one:

- [x] [Menu Manager](https://atom.io/packages/menu-manager) :arrow_right: https://github.com/jerone/menu-manager/pull/4
- [x] [Tool Bar Main](https://github.com/suda/toolbar-main/) :arrow_right: https://github.com/suda/toolbar-main/pull/15
- [ ] [Toolbar Almighty](https://github.com/varemenos/atom-toolbar-almighty) :arrow_right: https://github.com/varemenos/atom-toolbar-almighty/pull/15
- [x] [Particle Dev](https://github.com/spark/particle-dev) :arrow_right: https://github.com/spark/particle-dev/pull/150
- [x] [PlatformIO IDE](https://github.com/platformio/platformio-atom-ide) :arrow_right: https://github.com/platformio/platformio-atom-ide/pull/130
- [x] [Toolbar Shortcuts](https://github.com/JostCrow/atom-toolbar-shortcuts) :arrow_right: https://github.com/JostCrow/atom-toolbar-shortcuts/pull/5
- [x] [Tool Bar Atom](https://github.com/jeselxe/tool-bar-atom) :arrow_right: https://github.com/jeselxe/tool-bar-atom/pull/7
- [x] [Toolbar for Markdown-Writer](https://github.com/zhuochun/tool-bar-md-writer)
- [ ] [toolbar-basic](https://github.com/mattlovaglio/toolbar-basic) :arrow_right: https://github.com/mattlovaglio/toolbar-basic/pull/1
- [x] [atom-toolbar-config](https://github.com/MunGell/atom-toolbar-config) :arrow_right: https://github.com/MunGell/atom-toolbar-config/pull/1
- [ ] [atom-vital-tool-bar](https://github.com/taronpa/atom-vital-tool-bar) :arrow_right: https://github.com/taronpa/atom-vital-tool-bar/pull/1
- [x] [bloginatom](https://github.com/hevnly/bloginatom) :arrow_right: https://github.com/hevnly/bloginatom/pull/2
[...]

PR #153, opened 2016-07-13
:fire: Remove legacy spece-pen view api compat

zertosh:
Cleanup


package.json
```diff
@@ -35,7 +32,6 @@
     "tool-bar": {
       "description": "A container for buttons at workspace edge",
       "versions": {
-        "0.1.0": "provideToolBarLegacy",
         "1.0.0": "provideToolBar"
```
jerone: We could merge `0.1.0` and `1.0.0` together...

zertosh: Supporting `0.1.0` means having to keep `grim` and `jquery` :(

jerone: Not necessarily, in my [research](https://github.com/suda/tool-bar/issues/141) only 3 package used the extended API. Nuclide is up-to-date, Flex Tool Bar hopefully soon, and don't really care about Atom-macros.

zertosh: Maybe I'm misunderstanding what you meant by "merge". Are you saying:

```
"tool-bar": {
  "versions": {
    "0.1.0": "provideToolBar",    
    "1.0.0": "provideToolBar"
  }
}
```

That's kinda dangerous, since you don't know what kinds of private packages people have

jerone: Fair enough.

PR #160, opened 2016-08-30
Update One themes

simurai:
The One themes now handle the borders/padding of panels, so tool-bar doesn't have to worry about it. Thus this PR removes them. Ref:

- https://github.com/atom/one-light-ui/pull/69
- https://github.com/atom/one-dark-ui/pull/162

Also, it adds the same `.disabled` styles from One dark to One light.

The changes will land in Atom `1.11`.

### DIFF
--- a/styles/theme-one-dark-ui.less
+++ b/styles/theme-one-dark-ui.less
@@ -5,18 +5,4 @@
       background-image: inherit;
     }
   }
-  .tool-bar-top {
-    border-top: 0 none;
-  }
-  .tool-bar-right {
-    border-right: 0 none;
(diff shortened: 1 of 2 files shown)


jerone:
Will this have a lot of effect on current users (e.g. Atom 1.9.9 users), if I merge this and release the package later today?

simurai:
Not a big effect.. Just that there are extra borders top and right:

screen shot 2016-08-31 at 9 23 09 am

You can also easily wait for Atom `1.12` or `1.13`. Visually it looks the same (with or without this PR). So you don't have to worry about being late.

PR #162, opened 2016-09-11
Clean up

jerone:
Attempt to clean up Tool Bar package.

:art: :racehorse: :fire: 

@zertosh Feel free to chip-in (edits form maintainers is allowed).


lib/tool-bar-view.js
```diff
@@ -172,13 +172,12 @@ export default class ToolBarView {
         break;
     }
 
-    const classNames = [`tool-bar-${position.toLowerCase()}`];
```
zertosh: Adding all of the classes in one go is an optimization (albeit a small one)

jerone: For only two classes I found readability more important.

PR #163, opened 2016-09-14
✨ Plugin Manager

### LINKED ISSUE #49: Manage third-party buttons

jerone:
Closes #49 & https://github.com/suda/tool-bar/pull/132

Plugin Manager

Some comments:
* Config settings are set programmatically.
* New config settings (by installing package) require an Atom restart (Settings Views cache needs to be cleared for the setting to appear).
* By default all plugins are enabled (backwards compatible).
* Disabling & enabling works instantly (see screenie).
* All plugins settings in the Settings View are grouped in a section named "Plugins".
* Groups can't have description, but I think the plugin descriptions are sufficient.
* I don't think we can remove the config setting when plugin is uninstalled ?!
* Ideas "copied" from [Minimap](https://github.com/atom-minimap/minimap/blob/master/lib/mixins/plugin-management.js) package.
* This opens support for [custom button types](https://github.com/suda/tool-bar/issues/43).

TODO:
* [x] Implement plugin manager.
* [ ] Specs.
* [ ] Documentation.
* [ ] Remove consoles.
* [ ] Test, test, test.
* [ ] Squash commits.


spec/tool-bar-spec.js
```diff
@@ -76,9 +115,8 @@ describe('Tool Bar package', () => {
         toolBar = workspaceElement.querySelector('.tool-bar');
       });
 
-      it('by third-party packages', () => {
-        expect(toolBarAPI).toBeDefined();
-        expect(toolBarAPI.group).toBe('specs-tool-bar');
+      afterEach(() => {
+        toolBarAPI.unregister();
```
jerone: This might be redundant, as it appears the specs is deactivating the package after each run.

lib/tool-bar-plugin-manager.js
```diff
@@ -0,0 +1,61 @@
...
+    console.debug('ToolBarPluginManager::register', arguments);
+    let settingsKey = `tool-bar.plugins.${name}`;
+
+    const plugin = this.plugins[name] = new ToolBarManager(name, toolBar);
+    this.subscriptions[name] = new CompositeDisposable();
+
+    this.config.plugins[name] = {
+      type: 'boolean',
+      title: name,
+      description: `Whether the ${name} plugin is allowed to add items to the Tool Bar.`,
```
jerone: Description isn't visible in the Settings anymore. Atom must have changed something internal. 
Possible solution: `atom.config.setSchema`.

jerone: Fixed

spec/tool-bar-spec.js
jerone: And it does. IIRC/Jasmine calls `atom.reset`, which calls `@packages.reset()`, which in turn calls `@deactivatePackages()`.

jerone:
I've been running this PR for a few weeks now and I'm quite happy with it.

@suda @cakecatz @zertosh I would love if you guys could review and test before merging...

zertosh:
Looking

zertosh:
Oh man, there are a lot of perf killers in here, that are only going to make things worse. Here's a CPU profile of what happens during init while `tool-bar` is adding buttons:

tool-bar-cpu-profile

As you can see, for just 13 buttons (and 2 spacers), `tool-bar` is taking +100ms. Unfortunately this doesn't show up in `time-cop` as being attributed to `tool-bar`- but rather to the individual packages that added the buttons. Atom tallies up service consumption under `activationTime` for the consumer, not the provider. However, a quick profile shows you where the work is happening.

The reason this is happening is because DOM ops happen after every button is added. The way "Plugin Manager" is implemented, makes you still pay the price of adding a button, but then double penalizes you by touching the DOM to take it out. To make matters worse, explicitly writing to the config (i.e. `atom.config.set`/`atom.config.unset`) is *very* expensive (though these are batched by Atom, and run after packages load - but still).

IMHO, if a breaking change like Plugin Manager is going to be implemented, it makes sense to also add some sort of DOM queue (along with other DOM optimizations).

jerone:
Thank you so much for your reply. I really have to learn more about profiling. 

How does this PR compare, performance-wise, to the current situation? I don't want to introduce new performance killers, but this PR isn't focused about fixing existing ones.

You opened my eyes about the penalty by adding a button and then removing again. Will have a look how to improve that.

`atom.config.set` only happens once, when the plugin is for the first time is registered. A small overhead to register a plugin is expected.
`atom.config.unset` only happens when a plugin calls destroy. Clean-up when a plugin requests this I belief should be done.

This change is fully backwards compatible, so I don't understand your comment about "breaking change"...

Do you have an example of DOM queue solution?

This PR is open for maintainers, so if you want and have the time I would value your contribution.

cakecatz:
@jerone 👍 

Current disabling of plugin is not working at next running.
I think good if `plugin disabling` will disable package actually.

sep-29-2016 06-04-34

jerone:
@cakecatz An unchecked plugin checkbox should disable the plugin immediately and still at next startup. Can you confirm that your config file contains a key `tool-bar.plugins.<plugin>` with the value `false`...

jerone:
> I think good if plugin disabling will disable package actually.

I don't think that's a good idea. Some packages (e.g. Nuclide and menu-manager) provide button(s) as an extra service, while their primary goal is doing something different.

zertosh:
> How does this PR compare, performance-wise, to the current situation?

If a button is hidden, then you're increasing its cost because it gets added to the DOM first, then removed.

> atom.config.set only happens once, when the plugin is for the first time is registered.

So perf aside on this one, I know many people (including myself) that version control `~/.atom/config.cson`. This going to cause churn :( From looking at the code, I can guess that you're doing this because immediately below you call `atom.config.observe` and that was giving you `undefined`?

You can't access the defaults for exported configs during `load`. We've had that problem too. If you know what the default is supposed to be, then use that if you get `undefined` from `atom.config.get` (that means the user hasn't set a value). See https://github.com/facebook/nuclide/blob/8c8206d/lib/main.js#L188-L194.

> This change is fully backwards compatible, so I don't understand your comment about "breaking change"...

At least Nuclide assumes that buttons always have a parent node. If a user chooses to hide one, that's no longer true. This is fine, but def a breaking change :)

So now would be the time to implement the queue since that implies that consumers can't assume that a button in the DOM. I think the current implementation of enabled/disabled living as a separate, yet tightly coupled module, `tool-bar-plugin-manager.js`, makes the queue more complicated. I generally really like the idea of the user having control over the buttons, but maybe that logic belongs in `tool-bar-manager.js`. I have to think about it.

> Do you have an example of DOM queue solution?

Let me think about it, because the tricky part is only applying the queue during Atom's init. Or always applying it, but flush it on a `requestAnimationFrame`. I don't know what the profile of that is. I'll fiddle with it tomorrow and put up a RFC PR.

> This PR is open for maintainers, so if you want and have the time I would value your contribution.

Thank you :) I'll give this time tomorrow!

>> I think good if plugin disabling will disable package actually.

> I don't think that's a good idea.

Plus, group name != package name. You really have no way of knowing that name of the package that's adding a group.

jerone:
[**@zertosh**](https://github.com/zertosh) commented on [29 sep. 2016 05:09 CEST](https://github.com/suda/tool-bar/pull/163#issuecomment-250359768 "2016-09-29T03:09:51Z - Replied by Github Reply Comments"):
> > How does this PR compare, performance-wise, to the current situation?
> 
> If a button is hidden, then you're increasing its cost because it gets added to the DOM first, then removed.

Oh sorry, I meant other then adding-and-then-removing-issue (fully agree on that, am looking into that)...

> > atom.config.set only happens once, when the plugin is for the first time is registered.
> 
> So perf aside on this one, I know many people (including myself) that version control `~/.atom/config.cson`. This going to cause churn :( 

Actually I'm [syncing](https://atom.io/packages/sync-settings) [my config](https://gist.github.com/jerone/c4a69b06920a0471e71a) too.

> From looking at the code, I can guess that you're doing this because immediately below you call `atom.config.observe` and that was giving you `undefined`?

> You can't access the defaults for exported configs during `load`. We've had that problem too. If you know what the default is supposed to be, then use that if you get `undefined` from `atom.config.get` (that means the user hasn't set a value). See https://github.com/facebook/nuclide/blob/8c8206d/lib/main.js#L188-L194.

Thinking about it, I could probably remove [that block](https://github.com/suda/tool-bar/pull/163/files#diff-54907223f3ec35e247aaa95a941346aeR28) as the [default](https://github.com/suda/tool-bar/pull/163/files#diff-54907223f3ec35e247aaa95a941346aeR24) is already `true`. The idea was copied [from Minimap](https://github.com/atom-minimap/minimap/blob/master/lib/mixins/plugin-management.js#L232-L234). I have to check if it's not required for the dynamic config schema.

> > This change is fully backwards compatible, so I don't understand your comment about "breaking change"...
> 
> At least Nuclide assumes that buttons always have a parent node. If a user chooses to hide one, that's no longer true. This is fine, but def a breaking change :)

In an early draft the buttons where hidden with `display: none` for this reason. I removed that solution; because it was just to easy to force show the button again. Right now they are [indeed removed from their parent](https://github.com/suda/tool-bar/pull/163/files#diff-1a1d06dd483f0b6084823a52bf18e9fbR104). I've not seen any other package having issue with that. Would something break for Nuclide if you've build an extra parentnode check around it? 

> So now would be the time to implement the queue since that implies that consumers can't assume that a button in the DOM. I think the current implementation of enabled/disabled living as a separate, yet tightly coupled module, `tool-bar-plugin-manager.js`, makes the queue more complicated. I generally really like the idea of the user having control over the buttons, but maybe that logic belongs in `tool-bar-manager.js`. I have to think about it.

:+1:

> > Do you have an example of DOM queue solution?
> 
> Let me think about it, because the tricky part is only applying the queue during Atom's init. Or always applying it, but flush it on a `requestAnimationFrame`. I don't know what the profile of that is. I'll fiddle with it tomorrow and put up a RFC PR.

> > This PR is open for maintainers, so if you want and have the time I would value your contribution.
> 
> Thank you :) I'll give this time tomorrow!

Really value your input :heart: (and others)!

zertosh:
Ok so I was thinking about this, and I really do think it makes sense to merge the functionality from `tool-bar-plugin-manager.js` into `tool-bar-view.js`. The two are so tightly coupled that they really should be one and the same.

jerone:
I'm not sure I can see the benefit from that. If anything I like to move more things from `tool-bar-view` to `tool-bar-manager` and from `tool-bar-manager` to `tool-bar-plugin-manager`.

This is how I see it:

1. The PluginManager takes care of registering and unregistering plugins. It's the public interface that plugins can use.
2. The ToolBarManager takes care of the plugin request to add/remove buttons/spacers at the correct index in the correct tool-bar. At one point I like to extend this class with [custom buttons](https://github.com/suda/tool-bar/issues/43) and [multiple tool-bars](https://github.com/suda/tool-bar/issues/80).
3. The toolBarView takes care of adding DOM elements to the tool-bar.

I've not worked-out everything, as this is an ongoing process and not set in stone.

Can you elaborate why you want to combine `tool-bar-plugin-manager.js` into `tool-bar-view.js`?

aminya:
Why is this closed?

jerone:
[**@aminya**](https://github.com/aminya) commented on [Nov 10, 2019, 7:07 AM GMT+1](https://github.com/suda/tool-bar/pull/163#issuecomment-552166700 "2019-11-10T06:07:59Z - Replied by Github Reply Comments"):
> Why is this closed?

Because I moved away from Atom and no one took over this PR in 3 years. 
If anyone is interested in continuing, the [branch](https://github.com/jerone/tool-bar/tree/features/%2349-plugins-manager) is *still* there.

aminya:
@suda could you merge this?

suda:
@aminya it's been a while so if you're willing to continue Jeronen's work on this feature, we'd love a fresh PR 😊 The branch probably needs rebasing and most importantly testing 👍🏻

aminya:
I created a fresh up-to-date branch with this PR being on wait. Let's see if I find time to work on this:
https://github.com/aminya/tool-bar/pull/1

ericcornelissen:
@aminya the URL you provided does not work, also judging by the structure of the URL it looks like you opened the Pull Request on your fork of the repository rather than to this repository :thinking:

jerone:
@suda I cloned my branch into [this repo](https://github.com/suda/tool-bar/tree/features/%2349-plugins-manager). This way other people can work on it, after I delete my fork. Feel free to delete the branch if I was out of line.

aminya:
@ericcornelissen Oh sorry. I removed as I can't work on this for near future.

PR #165, opened 2016-09-29
Housekeeping

zertosh:
While working on batching DOM ops, these things caught my eye, so I figured I'd factor them out since they're pretty safe and unrelated.

cc: @jerone

### DIFF
--- a/lib/tool-bar-view.js
+++ b/lib/tool-bar-view.js
@@ -2,8 +2,6 @@
 
 import {CompositeDisposable, Emitter} from 'atom';
 
-const supportFullWidth = typeof atom.workspace.addHeaderPanel === 'function';
-
 export default class ToolBarView {
 
   constructor () {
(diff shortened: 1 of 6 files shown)


lib/tool-bar-view.js
```diff
@@ -47,19 +45,14 @@ export default class ToolBarView {
         } else {
           this.hide();
         }
+      }),
+      atom.config.onDidChange('tool-bar.fullWidth', ({newValue, oldValue}) => {
```
jerone: Disposable

lib/tool-bar-view.js
```diff
@@ -2,8 +2,6 @@
 
 import {CompositeDisposable, Emitter} from 'atom';
 
-const supportFullWidth = typeof atom.workspace.addHeaderPanel === 'function';
```
jerone: :+1: I'm fine with removing this; it's been there since 1.6.0 Stable

lib/tool-bar-view.js
jerone: Disposable

jerone:
One small change, other then that I'm +1 on merging.

jerone:
I merged #162 & #161 which cleaned up some tings too, resulting in (small) merge conflict here.

lib/tool-bar-view.js
zertosh: Yeah, I moved it into the giant `this.subscriptions.add` above it. It's just hard to see unless you expand the entire file.

jerone: You're absolutely right.

zertosh:
Rebased and fixed conflicts. Will merge once the tests are done.

jerone:
@cakecatz The change in this PR to rename the private property `toolBar` to `toolBarView` results in an error at `flex-tool-bar`. I noticed [this commit from you](https://github.com/cakecatz/flex-toolbar/commit/42d35de2f874e289edf00b5c01fecdbd6ec686a0) needing that property. There was no issue attached to it, so I was wondering why you need that property? Maybe we can fix it differently.

PR #171, opened 2016-10-11
allow multiple commands in an array

### LINKED ISSUE #170: allow multiple commands in an array

UziTech:
fixes #170


jerone:
Looks solid, but it will require some specs.

PR #175, opened 2016-11-02
Display keybinding in the tooltip

vjeux:
A little known feature of Atom tooltip is that it supports displaying the keybinding directly inside of the tooltip if you pass a `keyBindingCommand` value as the atom command.

In Nuclide we're already giving the command so I figured it would be cool to display the keybinding as well :)

Let me know if it's something you'd consider! Thanks

Test Plan:

Open Nuclide, mouse over all the buttons, make sure you can see the keybinding for it.
image

### DIFF
--- a/lib/tool-bar-button-view.js
+++ b/lib/tool-bar-button-view.js
@@ -15,11 +15,12 @@ export default class ToolBarButtonView {
     this.group = group;
 
     if (options.tooltip) {
-      this.element.title = options.tooltip;
+      const callback = this.options.callback;
       this.subscriptions.add(
         atom.tooltips.add(this.element, {
           title: options.tooltip,
-          placement: getTooltipPlacement
+          placement: getTooltipPlacement,
+          keyBindingCommand: typeof callback === 'string' ? callback : null
         })
       );
     }
--- a/spec/tool-bar-spec.js
+++ b/spec/tool-bar-spec.js
@@ -1,6 +1,7 @@
 'use babel';
 
 /* eslint-env browser */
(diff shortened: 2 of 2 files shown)


lib/tool-bar-button-view.js
```diff
@@ -15,11 +15,12 @@ export default class ToolBarButtonView {
     this.group = group;
 
     if (options.tooltip) {
-      this.element.title = options.tooltip;
```
vjeux: Note: if there's a title attribute on the DOM node, Atom will use it and not respect either title or keyBindingCommand passed as attribute.

It's doing weird things with the title here: https://github.com/atom/atom/blob/master/src/tooltip.js#L349-L354

spec/tool-bar-spec.js
```diff
@@ -95,7 +95,13 @@ describe('Tool Bar package', () => {
           tooltip: 'About Atom'
         });
         expect(toolBar.children.length).toBe(1);
-        expect(toolBar.firstChild.dataset.originalTitle).toBe('About Atom');
+        const element = toolBar.firstChild;
```
vjeux: The way of testing those is based on atom: it's not super pretty but works
https://github.com/atom/atom/blob/master/spec/tooltip-manager-spec.coffee

vjeux:
Just fixed the lint errors

captbaritone:
This change looks good to me. Is there anything I could do to help get this merged?

Another possible improvement to this PR would be to handle the case where `this.options.callback` is an object and the default `''` property is a command string.

PR #177, opened 2016-11-20
Add support for selected buttons.

gsmcmullin:
This allows consumers to implement toggle buttons or radio groups.  See #43 

The logic of toggling the selection is left to the consumer.  For a most basic toggle button:
```js
export default {
  toolBar: null,
  toggleButton: null,

  consumeToolBar(getToolBar) {
    this.toolBar = getToolBar('my-toolbar');
    this.toggleButton = this.toolBar.addButton({
      icon: 'plug',
      callback: () => this.toggleButton.setSelected(!this.toggleButton.getSelected())
    });
}
```

### DIFF
--- a/lib/tool-bar-button-view.js
+++ b/lib/tool-bar-button-view.js
@@ -51,6 +51,18 @@ export default class ToolBarButtonView {
     }
   }
 
+  setSelected (selected) {
+    if (selected) {
+      this.element.classList.add('selected');
+    } else {
+      this.element.classList.remove('selected');
+    }
+  }
+
+  getSelected () {
+    return this.element.classList.contains('selected');
+  }
+
   destroy () {
     this.subscriptions.dispose();
     this.subscriptions = null;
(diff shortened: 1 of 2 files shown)


gsmcmullin:
This is almost a duplicate of #143.  Sorry I didn't see that.  It is unclear why that was closed.
I can write specs for this if needed, but I thought I would submit it for comment before putting in the effort.

lib/tool-bar-button-view.js
```diff
@@ -51,6 +51,18 @@ export default class ToolBarButtonView {
...
+  setSelected (selected) {
+    if (selected) {
+      this.element.classList.add('selected');
+    } else {
+      this.element.classList.remove('selected');
+    }
+  }
+
+  getSelected () {
+    return (` ${this.element.className} `).indexOf(' selected ') > 0;
```
UziTech: shouldn't this be

```js
  return this.element.classList.contains('selected');
```

gsmcmullin: Fixed

suda:
Thanks Gareth!

edreesen:
this PR leads to #219, needs rework

PR #181, opened 2017-01-16
Added atom-material-ui styling

SpectralKH:


jerone:
The styling for Atom Material UI is added in the theme: https://github.com/atom-material/atom-material-ui/blob/master/styles/packages/tool-bar.less

I suggest you send a PR there with altered styling.

PR #188, opened 2017-03-31
Updating Font Awesome to v4.7.0

gregorygenova:


iconsets/font-awesome/variables.less
```diff
@@ -1,17 +1,21 @@
 // Variables
 // --------------------------
 
-@fa-font-path:        "atom://tool-bar/iconsets/font-awesome";
+@fa-font-path:        "../fonts";
```
suda: This seems to break `font-awesome` as Atom fails to load the font files with `net::ERR_FILE_NOT_FOUND`. It does work with `atom://tool-bar/iconsets/font-awesome`

Issue #193, opened 2017-04-23
Default iconset handling


idleberg:
I just came across [this question](https://discuss.atom.io/t/flex-tool-bar-icon-setting/42039) on the Atom forums, where a user wondered why his settings for `tool-bar` did not work.

```cson
{
    type: "button"
    icon: "arrow-down"
    callback: 'core:save'
    tooltip: "save"
    iconset: "octicons" # this does not work
  }
```
The reason why this doesn't work is that Octicons are default and need (and should) not be specified explicitly. When it is specified, no icon is displayed. IMHO, specifying `octicons` and omitting the `iconset` setting altogether should have the same effect.

Pyragon:
Fixed in #194

PR #194, opened 2017-04-26
Fix manually setting octicons as iconset

Pyragon:
Very simple fix to #193


Pyragon:
Fixed my eslinter so hopefully no more failed checks :p

PR #195, opened 2017-04-26
Add ability to specify octicons

### LINKED ISSUE #193: Default iconset handling

Pyragon:
Add ability to specify 'octicons' as the iconset, instead of needing to leave it blank.

Fixes #193

### DIFF
--- a/lib/tool-bar-button-view.js
+++ b/lib/tool-bar-button-view.js
@@ -1,146 +1,146 @@
 'use babel';
 
 import {
-	CompositeDisposable
+  CompositeDisposable
 } from 'atom';
 
 let prevFocusedElm = null;
 
 export default class ToolBarButtonView {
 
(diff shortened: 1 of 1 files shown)


suda:
Thanks for doing this Cody! Could you revert the style changes and just have the functional change in this PR?

PR #197, opened 2017-04-26
Add option for custom elements

Pyragon:
Add option for custom elements.
Also commits previous pull request.

Inteded for SVG, but can be used for anything really...
SVG's color + width/height is automatically set.
Usage: 
image


UziTech:
This can be closed. HTML was add in #217

PR #199, opened 2017-07-05
Added example for multiple callbacks

pheraph:
Had to search the issue tracker to find out this is even possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,12 @@ export function consumeToolBar(getToolBar) {
     },
     data: 'foo'
   });
+  
+  // Calling multiple callbacks at once
+  toolBar.addButton({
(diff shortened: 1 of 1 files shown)


aminya:
@suda How can I create a sequential callback? When I use this example it seems it executes the 2nd command before the first one finishing.

suda:
Unfortunately, this doesn't provide a way to control the execution order. You might create a single callback that then runs other callbacks in a sequential manner.

aminya:
I wrote one example here, if you want to add it to the readme:
https://github.com/aminya/juno-toolbar-plus/blob/60039275a1f07813cb0506f30b598c854bc9c4e2/lib/juno-toolbar-plus.coffee#L34-40

PR #211, opened 2018-01-07
Add Touch Bar support

suda:
This PR adds support for MacBook Pro Touch Bar. Due to Touch Bar limitations, it can only display seven buttons but that should be enough for most cases.


HsuanTingLu:
can I disable this by some kind of setting?

suda:
@HsuanTingLu yes, there's an option to turn it off: https://github.com/suda/tool-bar/issues/213#issuecomment-356117931

PR #216, opened 2018-01-16
allow tooltip to be an object

### LINKED ISSUE #182: option for html in tooltip
It would be nice to allow html in the tooltip either with an option in the settings or per button

UziTech:
Allows the tooltip option to be a string or object

fixes #182 
fixes https://github.com/cakecatz/flex-toolbar/issues/96


UziTech:
looks like travis failed because macos build got canceled. @suda can you restart the macos builds?

UziTech:
done

suda:
🙌 Woohoo! Thanks!

PR #217, opened 2018-01-16
Text/html button

### LINKED ISSUE #215: Add text, html options
It would be nice to add text instead of an icon sometimes. HTML might be nice too.

UziTech:
Add text and html options

html is a boolean attribute to allow the text to be inserted as html

I also made the icon optional to allow for just text buttons

fixes #215 
fixes https://github.com/cakecatz/flex-toolbar/issues/111
fixes https://github.com/cakecatz/flex-toolbar/issues/120

```js
  // Adding text button
  toolBar.addButton({
    text: 'hello',
    callback: 'application:about'
  });

  // Text buttons can also have an icon
  toolBar.addButton({
    icon: 'octoface',
    text: 'hello',
    callback: 'application:about'
  });

  // Text buttons can be html
  toolBar.addButton({
    icon: 'octoface',
    text: '<b>BIG</b> button',
    html: true,
    callback: 'application:about'
  });
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ export function consumeToolBar(getToolBar) {
 The method `addButton` requires an object with at least the property `callback`. The
 `callback` must be an Atom command string, a custom callback function or an
 object where the keys are key modifiers (`alt`, `ctrl` or `shift`) and the
-values are commands or custom function (see [example](#example)).
+values are commands or custom functions (see [example](#example)).
 
 The remaining properties
 `tooltip` (default there is no tooltip),
--- a/README.md
+++ b/README.md
@@ -178,6 +178,27 @@ export function consumeToolBar(getToolBar) {
     priority: 10
   });
 
+  // Adding text button
+  toolBar.addButton({
+    text: 'hello',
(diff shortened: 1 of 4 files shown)


UziTech:
@suda any feedback on this?

UziTech:
@suda I was able to implement this in https://github.com/cakecatz/flex-toolbar/pull/125 but I had to use [`!important`](https://github.com/cakecatz/flex-toolbar/blob/36be07a669baa95b0eb06cc5b07d29f51f29ae24/styles/flex-tool-bar.less#L30) in the css to override the width. So it would be a lot better to implement it in tool-bar so I could remove that style hack.

suda:
Oh sorry for keep you waiting! Merging it now :)

edjubuh:
Does this close #197?

UziTech:
@edjubuh yes, you can add svg elements with:

```js
toolbar.addbutton({
	tooltip: 'Cherry Pick',
	text: '<svg ...',
	html: true,
	callback: 'git-plus:cherry-pick'
});
```

PR #218, opened 2018-01-31
adding color, no-gradient spacer, button sizes, update mdi

evehne:
1) adding button "color" & "background" option support. As CSS color (#, rgb, rgba ...) because unicorns and life is color. we do as we want.
2) adding buttons alternatives sizes options :
* 24px is good but make limits of btn numbers on screen width < 1920px
* 16px make eyes hurt if you not on 800x600 screen ... on windows 3.11 ... in 90's ! 
* 32px is for children ui or minimal button size for phone. PC/Mac ? 32px on toolbar is impratical.
So : 
- 18px, 21px, 28px better than 16px (too small) / 24px (too big) / 32px (too big)
- 21px is good & usefull size for main menu bar with lots of btns on wide/ultrawide screen (1920 / 2560)
- 28px is good on small screen with big res.
- 18px is good for make tiny tool bar without bleeding eyes.
3) adding no-gradient / transparent spacer. 
* remember this : _Space is the art of to create sufficient discernment without breaking continuity_. Space is the opposite of trump... space should not, never, be a wall between button !
- so transparent spacer made you and your ui better, without covfefe.
4) update MDI iconset to 2.1.19.


aminya:
Other than no-gradient space, I added color and more icon sizes in https://github.com/suda/tool-bar/pull/277, and https://github.com/suda/tool-bar/pull/278

I don't know if the spacer is something interesting to have

ericcornelissen:
> We can close this if we want...

I agree, also considering how outdated this Pull Request is. Agreed @suda?


> I don't know if the spacer is something interesting to have.

I think this would be a nice thing to have. as @suda pointed out "IMO the transparent spacer is a good idea but maybe more as something user configurable?" But we can just open a new issue for that.

aminya:
> > We can close this if we want...
> 
> I agree, also considering how outdated this Pull Request is. Agreed @suda?
> 
> > I don't know if the spacer is something interesting to have.
> 
> I think this would be a nice thing to have. as @suda pointed out "IMO the transparent spacer is a good idea but maybe more as something user configurable?" But we can just open a new issue for that.

I can make a PR for that one too. But I wasn't sure if it is wanted by the community.

suda:
Agreed. Closing this one in favor of Amin's PRs

PR #220, opened 2018-02-10
set this as button in callback function

UziTech:
This will allow the user to use `this.getSelected()` in the function callback


suda:
Thanks again Tony!

PR #239, opened 2018-08-27
Avoid taking focus when clicking buttons

### LINKED ISSUE #235: A simpler solution to the problem of focus?
TL;DR: _There's another bug with handling focus when clicking on a toolbar button. I think there may be a simpler solution._

Tool Bar tries to behave as if the command that the toolbar button triggers was triggered from whatever had focus before the user clicked. As a result, Tool Bar actively gives focus to the DOM node that was focused before the user moved their mouse into the toolbar. Unfortunately giving a nodes can have side effects. For example, a pane will reveal itself when it receives focus.

This causes a bug, which can be reproduces by:

1. Have two Panels: PanelA and PanelB which are both in a visible dock
2. Focus PanelA
3. Move mouse to the toolbar (prevFocusedElm is now PanelA)
4. Click button which toggle's PanelB (PanelB is now focused)
5. Click button which toggle's PanelB (The dock containing PanelA is hidden)
6. Click button which toggle's PanelB

What happens now is:

1. prevFocusedElm (which is still PanelA) gets focused, causing the dock (and PanelB) to become visible
2. The command which toggles PanelB is now called with PanelB visible, so it closes PanelB instead of opening it

The net effect is that the third (and any subsequent) clicks in the toolbar do nothing, and the feature seems broken.

Proposed Solution
 =================

Rather than taking focus and then attempting to return it, what if wen never took that focus in the first place?

By calling `e.preventDefault` in a `mousedown` handler, we can avoid taking focus, and therefore we never have to return it. We may even be able to avoid tracking `prevFocusedElm` all together.

**If this solution seems reasonable, I'm happy to open a pull request.**

I think @AnWeber alluded to this issue on a comment here back in 2016: https://github.com/suda/tool-bar/pull/88

captbaritone:
Rather than tracking the previously focused element, don't take
focus in the first place.

This avoids a few error cases:

1. Returning focus can have unintentional side effects.
2. Theoretically focus could change between the time a user moused into
   the tool bar and when they actually clicked.

Resolves #235

### DIFF
--- a/lib/tool-bar-button-view.js
+++ b/lib/tool-bar-button-view.js
@@ -1,7 +1,5 @@
 const {CompositeDisposable} = require('atom');
 
-let prevFocusedElm = null;
-
 module.exports = class ToolBarButtonView {
 
   constructor (options, group) {
@@ -60,11 +58,11 @@ module.exports = class ToolBarButtonView {
 
     this.element.classList.add(...classNames);
 
+    this._onMouseDown = this._onMouseDown.bind(this);
     this._onClick = this._onClick.bind(this);
-    this._onMouseOver = this._onMouseOver.bind(this);
 
+    this.element.addEventListener('mousedown', this._onMouseDown);
     this.element.addEventListener('click', this._onClick);
-    this.element.addEventListener('mouseover', this._onMouseOver);
   }
 
   setEnabled (enabled) {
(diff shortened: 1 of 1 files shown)


captbaritone:
@suda

captbaritone:
Awesome! Thanks for the quick turn around. 🎉

suda:
Np.! Just released to `apm`: https://github.com/suda/tool-bar/releases/tag/v1.1.8 :)

PR #240, opened 2018-08-27
Remove ineffective focus test

captbaritone:
The tool bar is never actually added to the DOM, so the `.focus()` call
does not actually do anything.

If this test were working as expected, it should have failed on https://github.com/suda/tool-bar/pull/239.


suda:
Thanks Jordan!

PR #243, opened 2018-08-30
Remove mousedown event listener

captbaritone:
Whoops. Overlooked this in #239.


suda:
Released: https://github.com/suda/tool-bar/releases/tag/v1.1.9 🙌🏻

Issue #250, opened 2018-11-05
Save button icon disappeard


Deduction42:
Using the toolbar with Julia+Juno, and when updating to version 1.1.11, the "save" button now has no icon. This happened in both Windows 10 and Linux Mint 19.

asbisen:
I am encountering this issue on macOS mojave too.

unluisco:
thanks for the report @Deduction42,

which plugin are you using with tool-bar?

jens-t:
> I looked at the specific icons having issues and they appear to have been removed or renamed in the latest version of fa(font awesome) icon set.

Thanks for checking, I assumed something like that, cause the icons that are still displayed also have a different style and are displayed in a bigger size:

![bildschirmfoto-20181107171619-80x448](https://user-images.githubusercontent.com/9143480/48144498-55b7d580-e2b1-11e8-81eb-d38736908493.png)

Is the size intentional? Seems a bit off, i thank the icons need more padding ... (In the Screenshot you see the 16px size setting)

unluisco:
I checked the icons, and yes, the terminal icon is a bit larger than normal, but that would be a matter of the [fontawesome](https://github.com/FortAwesome/Font-Awesome) designers.

jens-t:
> I checked the icons, and yes, the terminal icon is a bit larger than others, but that would be a matter of the fontawesome designers.

Ok, thanks! I might take a closer look at the other icon sets or adjust the icon size in the styles.less ...

asbisen:
@burneyhoel how did you change the icons to use the new names. Also how did you find the new names :) 

> Edit 2: I looked at the specific icons having issues and they appear to have been removed or renamed in the latest version of fa(font awesome) icon set. I changed my icons to the new names and the problem was resolved. This means it is not an issue with the tool bar package, the package that is using tool bar will need to update the icons used.

unluisco:
@asbisen which plugin are you using with tool-bar?

burneyhoel:
@asbisen In the package that uses tool-bar, edit the tool-bar-(package).coffee file for the items not working(example below). I found the new names on the website for font awesome and i changed a few to use different icon sets.

Supported icon sets
Octicons (Atom's flavour)
Ionicons (ion)
FontAwesome (fa)
Foundation (fi)
IcoMoon (icomoon)
Devicon (devicon)
MaterialDesignIcons (mdi)

example:
(OLD)
    @toolBar.addButton
      callback: 'settings-view:open'
[...]

PR #257, opened 2019-07-11
Update font-awesome to 5.9.0 including brand icons

malnvenshorn:
* Updated the font-awesome icon set to version 5.9.0
* brand icons are now accessible using `iconset: "fab"`

Relevant ticket: #250


suda:
Thanks for doing this Sven!

Issue #204, opened 2017-08-27
Status of the project


wdanilo:
Hi! I just want to ask if this project is alive or is it slowly dying due to lack of maintainers? There is a lot of important issues on the tracker and many important pull requests waiting to be merged, however, literally nothing happened to this project for almost the last year

suda:
Hi Wojtek! It is true, we haven't spent a lot of time recently on maintaining this package (I tried to clear up some PRs and issues in the past month). 

**We would definitely welcome any help though if you (or any other user) are interested!**

ericcornelissen:
I would be glad to help out, at least for a while :+1:

suda:
That would be much appreciated Eric! Thank you!

PR #258, opened 2019-08-12
Combine string and function callback for button view

ericcornelissen:
# Change

The change I'm proposing allows users to specify both atom commands and function for callback when multiple callbacks are specified. So, before it was only possible to do:

```js
toolBar.addButton({
  type: "button",
  icon: "chevron-right",
  callback: ['application:cmd-1', 'application:cmd-2'],
  tooltip: "Show project"
})
``` 

But not with a command and function, like:

```js
toolBar.addButton({
  type: "button",
  icon: "chevron-right",
  callback: ['application:cmd-1', () => console.log("Hello world!")],
  tooltip: "Show project"
})
``` 

With this Pull Request that is now possible.

# Implementation

I changed the "executeCallback" function so that the user can use both strings (for atom command dispatching) and function (for whatever the user wants) when the `callback` option is set to an array.

To achieve this I convert the `callback` value to an array if it is not and then iterate over the array and perform either the dispatch or function call as it happened before.

# Testing

I did not find any tests for this function so I also didn't add a new test.

### DIFF
--- a/lib/tool-bar-button-view.js
+++ b/lib/tool-bar-button-view.js
@@ -131,7 +131,7 @@ module.exports = class ToolBarButtonView {
       : workspaceView;
 
     if (!Array.isArray(callback)) {
-      callback = [callback]
+      callback = [callback];
     }
 
     for (let i = 0; i < callback.length; i++) {
--- a/lib/tool-bar-button-view.js
+++ b/lib/tool-bar-button-view.js
@@ -130,14 +130,16 @@ module.exports = class ToolBarButtonView {
       ? document.activeElement
       : workspaceView;
 
-    if (typeof callback === 'string') {
-      atom.commands.dispatch(target, callback);
-    } else if (Array.isArray(callback)) {
-      for (let i = 0; i < callback.length; i++) {
+    if (!Array.isArray(callback)) {
+      callback = [callback]
+    }
+
+    for (let i = 0; i < callback.length; i++) {
+      if (typeof callback[i] === 'string') {
         atom.commands.dispatch(target, callback[i]);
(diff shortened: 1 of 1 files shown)


suda:
Thanks Eric! I see the tests fail on Beta channel for MacOS: https://travis-ci.org/suda/tool-bar/jobs/570771960 I guess that's Atom's fault. If you have a Mac, could you try running them and see it that's the case?

ericcornelissen:
@suda, looking at  the error logs [it say the following for every failing test](https://travis-ci.org/suda/tool-bar/jobs/570771960#L112):

```
Expected promise to be resolved, but it was rejected with: Cannot read property 'provideToolBar' of null {  }
```

This is, I'm guessing, referring to this (the only other `provideToolBar` is [here](https://github.com/suda/tool-bar/blob/e1065dbf1aac8e6dfee964758d386deebd5ecb98/lib/tool-bar.js#L20)):

https://github.com/suda/tool-bar/blob/e1065dbf1aac8e6dfee964758d386deebd5ecb98/spec/tool-bar-spec.js#L35-L38

So, it seems to me that the build that's failing had some problems activating the package (can't imagine that is due to my changes). Rerunning it might make the build work again.

<sup>Either way I don't have a Mac to test this on </sup>

ericcornelissen:
Anything that is preventing this from being merged @suda? If it is the failiing build I (again) suggest to restart the build.

Also, as [I said earlier](https://github.com/suda/tool-bar/issues/204#issuecomment-520367813) I'm will to help out maintaining the package 😉

suda:
Oh sorry, Eric! This got buried under other stuff 😓 Merging it in 👍

PR #260, opened 2019-09-07
separators look better with dark themes

gilbertohasnofb:
With dark themes, such as the One Dark UI, the separators are virtually invisible. These values I used seem to work well with both dark and light UI's.

### DIFF
--- a/styles/tool-bar.less
+++ b/styles/tool-bar.less
@@ -47,15 +47,15 @@
 
   &.tool-bar-horizontal .tool-bar-spacer {
     border: 0 none;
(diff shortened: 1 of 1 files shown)


gilbertohasnofb:
I am closing this because further testing shows that the issue is only with the One Dark and One Light UI's. All others are fine. This PR actually make the former two look better, but all others look worse. I am opening an issue instead as this is not a good solution.

gilbertohasnofb:
See: https://github.com/suda/tool-bar/issues/261

PR #251, opened 2018-11-06
update ionicons icons (release 4.4.6)

luiscobits:
the test is waiting for the ionic icon to be `f14b` but now it is `f150`.
because of that, the test fail.

PR #263, opened 2019-10-01
Update ionicons icons (release v4.5.10-1)

ericcornelissen:
Based on #251 I decided to update the ionicons iconpack to the latest version. I'm not sure why the test was failing there, for me it was simply the case that the css classname for the icon has changed.

@suda, have you ever considered including the iconpacks as dependencies, that would make it possible to update them more quickly/easily/automatically.


suda:
Thanks Eric! I believe that Atom couldn't resolve the path to the font files when being imported from the CSS when using npm dependencies, but feel free to try, maybe this isn't true anymore.

UziTech:
Looks like this update removed a bunch of icons from the ion set.

`ion-refresh` and `ion-close` to name a few

UziTech:
Looks like it removed all material style icons and only left ios style icons

UziTech:
or material design icon now need to be prefaced with `md-`

Mayccoll:
Right the `md-` fix the issue.
Thanks for the quick response

ericcornelissen:
> or material design icon now need to be prefaced with `md-`

This seems to be what happened :wink:

PR #267, opened 2019-10-03
Update icons

UziTech:
Updates the following icon lists:
- mdi v4.4.95
- font-awesome v5.11.2
- devicons v2.2

I'm fairly sure none of these updates are breaking changes. Just adding new icons not changing old names.


suda:
Thanks Tony!

PR #277, opened 2020-02-16
Allowing more iconsizes

aminya:
- Adds icons size of 18,21,28
- Setting default size to 21

Derived from https://github.com/suda/tool-bar/pull/218

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,10 +32,13 @@ On which edge of the editor should the tool bar appear. Possible options:
 
 ### Icon size
 
-*   Very Small *(12px)*
-*   Small *(16px)*
+*   Tiny *(12px)*
+*   Very Small *(16px)*
+*   Small: *(18px)*
+*   Medium: *(21px)*
 *   Big *(24px)*
-*   Large *(32px)*
+*   Very Big *(28px)*
+*   Huge *(32px)*
 
 ### Visibility
 
--- a/lib/tool-bar-view.js
+++ b/lib/tool-bar-view.js
@@ -4,11 +4,6 @@ const rafDebounce = require('./raf-debounce');
 module.exports = class ToolBarView {
 
   constructor () {
-    // warning for Atom-Material-UI
-    if (atom.themes.getEnabledThemeNames()[1] === 'atom-material-ui') {
-      atom.notifications.addWarning('Using "Atom-Material-UI" theme with "Tool-Bar" is not recommended. Use other themes such as "One-Datk-UI".');
-    };
-
     this.element = document.createElement('div');
     this.element.classList.add('tool-bar');
     this.items = [];
--- a/lib/tool-bar-view.js
+++ b/lib/tool-bar-view.js
@@ -128,7 +128,10 @@ module.exports = class ToolBarView {
     this.element.classList.remove(
       'tool-bar-12px',
       'tool-bar-16px',
+      'tool-bar-18px',
+      'tool-bar-21px',
       'tool-bar-24px',
+      'tool-bar-28px',
       'tool-bar-32px'
     );
     this.element.classList.add(`tool-bar-${size}`);
--- a/lib/tool-bar-view.js
+++ b/lib/tool-bar-view.js
@@ -4,6 +4,11 @@ const rafDebounce = require('./raf-debounce');
 module.exports = class ToolBarView {
 
   constructor () {
+    // warning for Atom-Material-UI
+    if (atom.themes.getEnabledThemeNames()[1] === 'atom-material-ui') {
+      atom.notifications.addWarning('Using "Atom-Material-UI" theme with "Tool-Bar" is not recommended. Use other themes such as "One-Datk-UI".');
+    };
+
     this.element = document.createElement('div');
     this.element.classList.add('tool-bar');
     this.items = [];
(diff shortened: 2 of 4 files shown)


aminya:
Could you review this? @suda

README.md
```diff
@@ -33,8 +33,11 @@ On which edge of the editor should the tool bar appear. Possible options:
 ### Icon size
 
 *   Very Small *(12px)*
-*   Small *(16px)*
+*   Smaller *(16px)*
+*   Small: *(18px)*
```
ericcornelissen: Perhaps change this to :arrow_down:?

```markdown
*   Tiny *(12px)*
*   Vary small *(16px)*
*   Small: *(18px)*
```

As far as I know changing these names doesn't break backwards compatibility.

lib/tool-bar.js
```diff
@@ -29,8 +29,8 @@ exports.config = {
   },
   iconSize: {
     type: 'string',
-    default: '24px',
-    enum: ['12px', '16px', '24px', '32px'],
+    default: '21px',
```
ericcornelissen: I'm personally not sure if changing the default is a good idea (also given my comment above).

aminya:
> 
> 
> Hi @aminya, I'm not a maintainer but I'm just trying to help @suda out with reviews.
> 
> The implementation looks good in theory, but for some reason I'm getting these weird "blocks" around my buttons **only** for the new sizes you added. See an example below: left is 21px, right is 24px 🤔 Do you have them too? Any ideas what might be causing them?
> 
> preview
> 
> I'm using:
> 
>     * UI theme: [atom-material-ui](https://atom.io/themes/atom-material-ui)
> 
>     * Syntax: [one-dark-material-syntax](https://atom.io/themes/one-dark-material-syntax)
> 
>     * Toolbar: [flex-tool-bar](https://atom.io/packages/flex-tool-bar)

I don't have these. Maybe this problem. I tried with different themes, and it was fine.

README.md
aminya: Fixed in https://github.com/suda/tool-bar/pull/277/commits/3489f07427c703faaf7f2425a6a41c94dc82923d

lib/tool-bar.js
aminya: I don't feel strongly about it. We can return to 24. Whatever you guys think.

ericcornelissen:
Thanks for the response @aminya, I investigated further and it turns out the discrepancy is caused by [atom-material-ui](https://atom.io/themes/atom-material-ui) (it removes margins for existing button sizes but not for the new button sizes) so you don't need to worry about it 👍 


> Your photo doesn't look bad though...

Not too bad, but definitely worse in my opinion, but that is objective. Subjectively speaking, it is very weird that the `21px` buttons take up **more** space than the `24px` buttons.

aminya:
> 
> 
> Thanks for the response @aminya, I investigated further and it turns out the discrepancy is caused by [atom-material-ui](https://atom.io/themes/atom-material-ui) (it removes margins for existing button sizes but not for the new button sizes) so you don't need to worry about it 👍
> 
> > Your photo doesn't look bad though...
> 
> Not too bad, but definitely worse in my opinion, but that is objective. Subjectively speaking, it is very weird that the `21px` buttons take up **more** space than the `24px` buttons.


Atom material looks very bulky for me. It has more problems that make it the worse theme for me. 
- It doesn't show the tab names correctly (just squashed together)
- The project management is large and bulky for no reason
- The settings seem just cluttered
- New one: It doesn't show 21px toolbar buttons correctly.

I was never able to use it, and now it is even worse. 

I love One-Dark.

Compare One-Dark:
image

To Material:
image

aminya:
They don't pay attention to the issues: https://github.com/atom-material/atom-material-ui/issues/483, so I guess we should just ignore the problem with Meterial

ericcornelissen:
> They don't pay attention to the issues: [atom-material/atom-material-ui#483](https://github.com/atom-material/atom-material-ui/issues/483), so I guess we should just ignore the problem with Material

I don't think we should it into account in the context of this Pull Request.


> I can add a warning about the issues with Material UI to the readme, or even a simple code that detects if the user is using Atom-UI-Meterial and giving warning about that.

Not sure if that is needed, I personally don't think so. If you think it is very easy to do you can share some code that users can put in their personal scripts/stylesheets to fix the problem locally?

aminya:
> 
> Not sure if that is needed, I personally don't think so. If you think it is very easy to do you can share some code that users can put in their personal scripts/stylesheets to fix the problem locally?

Yes the warning code is simple. I added the it.

lib/tool-bar.js
suda: I'd definitely keep the 24px default. There's no reason for changing it.

lib/tool-bar-view.js
```diff
@@ -4,6 +4,11 @@ const rafDebounce = require('./raf-debounce');
 module.exports = class ToolBarView {
 
   constructor () {
+    // warning for Atom-Material-UI
+    if (atom.themes.getEnabledThemeNames()[1] === 'atom-material-ui') {
```
suda: I'm not fond of this check. Can we have a CSS override for `atom-material-ui` in this package if they are not willing to fix it?

aminya:
I addressed the review comments. This is now ready to get merged.

PR #281, opened 2020-02-23
Fix build error for #272

ericcornelissen:
This Pull Requests merges fixes for the build errors in #272 into #272


suda:
Thanks Eric!

PR #286, opened 2020-03-06
Update plugins in the readme

aminya:
* Adding Juno Plus to plugins (TypeScript example)
* Move archived packages lower

If there are better plugins, we should add them to the list and remove the unmaintained packages.


ericcornelissen:
Seems like a nice addition to the list, at least in my opinion. I've seen that toolbar come by a couple of times in issues so its at least somewhat popular :+1: 

While we're looking at this list, I'd figured we might as well do some tidying up so I created #287 :smiley:

suda:
Thanks Amin! I merged some other PRs so you might need to rebase yours :)

aminya:
I rebased the PR. This is now ready to get merged.

PR #289, opened 2020-04-01
Type declarations for TypeScript

aminya:
This adds types for tool-bar public API. 

However, since this package is not an `npm` package, we should either push this to [@types/atom](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/atom)
or make a separate `@types/tool-bar`. 

I prefer the first option, `../index` is refering to that file.

Either way, it is good to keep a copy of that file here.

See this in action here: https://github.com/aminya/juno-plus

### DIFF
--- a/@types/index.d.ts
+++ b/@types/index.d.ts
@@ -153,3 +153,26 @@ export declare class ToolBarManager {
     /** The onDidDestroy method takes a function that will be called when the tool-bar package is destroyed. This is useful if your package needs to do cleanup when the tool-bar is deactivated but your package continues running. */
     onDidDestroy(callback: () => void): void
 }
+
+/**
+ *  Passed as an input to `consumeToolBar(getToolBar: getToolbarCallback)` function of your package.
+ *
+ *  In your main package file, add the following methods and replace your-package-name with your package name.
+ * ```js
+ *  let toolBar: ToolBarManager
+ *
+ *  export function consumeToolBar(getToolBar: getToolbarCallback) {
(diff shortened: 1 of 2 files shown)


aminya:
If you agree with the content of the docstrings and the declarations itself, I will make a pull request to `@types/atom`.

@types/index.d.ts
```diff
@@ -0,0 +1,185 @@
...
+    /** Adds a button. The input to this function is a `ButtonOptions` object */
+    addButton(options: ButtonOptions): void
+
+    /** Adds a spacer. Optionally, you can pass a `SpacerOptions` object */
+    addSpacer(options?: SpacerOptions): void
+
+    /** Use the method removeItems to remove the buttons added by your package. This is particular useful in your package deactivate method, but can be used at any time. */
+    removeItems(): void
+
+    /** The onDidDestroy method takes a function that will be called when the tool-bar package is destroyed. This is useful if your package needs to do cleanup when the tool-bar is deactivated but your package continues running. */
```
ericcornelissen, line 160: Maybe wrap this documentation to reduce the line length?

aminya: Done in https://github.com/suda/tool-bar/pull/289/commits/5da3bf1d82a54254edb2f177b0f0c13d45dab92c

aminya:
@suda Could you review this?

suda:
Thanks Amin!

aminya:
PR is created: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43727

Issue #291, opened 2020-04-08
Passing HTML element to addButtonByElement


aminya:
# Proposal
Similar to Status-Bar that has a function that accepts an HTML element, we should allow this too. This feature in combination with the typical `addButton` allows complex usage of Toolbar.

```ts
toolbar.addButtonByElement(element)

```
Then, the user can do whatever they wanna do with it. For example, define custom buttons like dropdowns, define click callbacks, hide it when they want, etc. It easily solves https://github.com/suda/tool-bar/issues/43 and https://github.com/suda/tool-bar/issues/192 by removing the limitations.

# Implementation
**This implementation is old. See the actual PR: https://github.com/suda/tool-bar/pull/294/files**

we just need to define an `addButtonByElement`:
In https://github.com/suda/tool-bar/blob/master/lib/tool-bar-manager.js#L11-L16
```ts
  addButtonByElement(element) {
    this.toolBarView.addElement(element);
    this.touchBarManager.addElement(element);
    return element;
  }
```

and define `toolBarView.addElement` which is edited from `addItem`.
https://github.com/suda/tool-bar/blob/0b8a1ced016b852ed175f72d7a860b324ef438fd/lib/tool-bar-view.js#L65-L90

Similarly, `touchBarManager.addElement` can also be defined.

# Alternative implementation
Instead of defining a new function, we can use the curren `addButton` and get the element as a property (named element or something), and instead of [creating the element inside ToolBarButtonView](https://github.com/suda/tool-bar/blob/0b8a1ced016b852ed175f72d7a860b324ef438fd/lib/tool-bar-button-view.js#L5), use the supplied element. User can use some desired features of `addButton` regularly (if their element is compatible). `callback` property should be optional in this case (we need to add `if` for all of the options in [`ToolBarButtonView`](https://github.com/suda/tool-bar/blob/0b8a1ced016b852ed175f72d7a860b324ef438fd/lib/tool-bar-button-view.js#L3). Even for [`onClick`](https://github.com/suda/tool-bar/blob/0b8a1ced016b852ed175f72d7a860b324ef438fd/lib/tool-bar-button-view.js#L82))
```ts
toolbar.addButton({
 element: myHTMLElement
 icon: "play"
 // etc
})
```
This implementation is easier to work with, since a lot of options are already available to use.

# Usage Example

This could be a typical example of that button (similar to https://github.com/aminya/atom-indent-detective/blob/TypeScript/src/status.ts)
```ts
// Called only once for each Atom session
[...]

Issue #192, opened 2017-04-18
[Feature Request] Scope based buttons


ethernal:
Can you add ability to limit buttons visibility based on scope ex. text.md could then have buttons specific to markdown editing, but the same buttons are not required when editing cpp or other file types.
If undefined button should be visible in all scopes - that way you don't introduce BC and hard requirement on scope definition.

Re-posting this from: https://github.com/varemenos/atom-toolbar-almighty/issues/24

PR #294, opened 2020-04-12
Add button by element

aminya:
Fixes https://github.com/suda/tool-bar/issues/291

# Motivation
Similar to Status-Bar that has a function that accepts an HTML element, we should allow this too. This feature in combination with the typical `addButton` allows complex usage of Toolbar.

```ts
item = toolbar.addItem({
  element: myElement,
  priority: someNumber // optional
})
```

Then, the user can do whatever they wanna do with it. For example, define custom buttons like dropdowns, define click callbacks, hide it when they want, etc. It easily solves https://github.com/suda/tool-bar/issues/43 and https://github.com/suda/tool-bar/issues/192 by removing the limitations.

# Changes
- Defined ToolBarItem which is the parent (super) of all classes (basically refractors the common part of ToolBarButton and ToolBarSpacer)
- addItem function that constructs a ToolBarItem and adds it to the Toolbar.
- Made ToolBarButton inherit from ToolBarItem 
- Made ToolBarSpacer inherit from ToolBarItem 
- Made classes named exports


suda:
Nice! LGTM :) If you add some docs to the readme, we could ship it! Do you agree @ericcornelissen ?

ericcornelissen:
Implementation looks good to me :+1: (not sure if the comment about calling super in the constructor are necessary, as that's pretty common for classes :thinking:)

aminya:
I have a question regarding the `consumeToolBar`. Is there a way to export `ToolBarItem` using the service? 

It seems that since `apm` isn't like `npm` people will have difficulty using the `ToolBarItem` class for expanding on it. Unless you think it is possible.

ericcornelissen:
Small detail, but I would prefer to move the `addItem` documentation after the `addSpacer` documentation as `addItem` is the "most advanced" method (or at least after `addButton`, in alphabetical order).

aminya:
@ericcornelissen done in https://github.com/suda/tool-bar/pull/294/commits/533d8f60941904ea76dabe85bb5c669349ae32d4

PR #299, opened 2020-04-21
Less compiling solution

aminya:
Solves the less compiling problem.

Fixes how spacers look like:
image

Compared to now color for spacers before this:

### DIFF
--- a/styles/tool-bar.less
+++ b/styles/tool-bar.less
@@ -38,6 +38,7 @@
     }
   }
 
+  // Set the margin between the buttons (that are not at the end)
   &.tool-bar-horizontal .tool-bar-btn:not(.tool-bar-item-align-end) + .tool-bar-btn {
     margin-left: 0;
   }
@@ -61,8 +62,7 @@
   }
 
   &.tool-bar-horizontal {
-    .tool-bar-btn,
-    .tool-bar-spacer {
+    .tool-bar-btn, .tool-bar-spacer {
       &.tool-bar-item-align-end {
         order: 2;
         &:first-child {
@@ -73,8 +73,7 @@
   }
 
   &.tool-bar-vertical {
-    .tool-bar-btn,
-    .tool-bar-spacer {
+    .tool-bar-btn, .tool-bar-spacer {
       &.tool-bar-item-align-end {
         order: 2;
         &:first-child {
--- a/styles/ui-variables.less
+++ b/styles/ui-variables.less
@@ -1,0 +1,85 @@
+// This file has fallback variables. It specifies the variables themes
+// must implement.
+
+// Colors
+
+@text-color: #333;
+@text-color-subtle: #777;
+@text-color-highlight: #111;
+@text-color-selected: @text-color-highlight;
+
+@text-color-info: #5293d8;
+@text-color-success: #1fe977;
+@text-color-warning: #f78a46;
+@text-color-error: #c00;
+
+@background-color-info: #0098ff;
+@background-color-success: #17ca65;
+@background-color-warning: #ff4800;
+@background-color-error: #c00;
+@background-color-highlight: hsla(0,0%,0%,.1);
+@background-color-selected: @background-color-highlight;
+
+@app-background-color: #fff;
+
+@base-background-color: #fff;
+@base-border-color: #eee;
+
+@pane-item-background-color: @base-background-color;
+@pane-item-border-color: @base-border-color;
+
+@input-background-color: #fff;
+@input-border-color: @base-border-color;
+
+@tool-panel-background-color: #f4f4f4;
+@tool-panel-border-color: @base-border-color;
+
+@inset-panel-background-color: #eee;
+@inset-panel-border-color: @base-border-color;
+
+@panel-heading-background-color: #ddd;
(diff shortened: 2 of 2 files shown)


styles/tool-bar.less
```diff
@@ -61,8 +62,7 @@
   }
 
   &.tool-bar-horizontal {
-    .tool-bar-btn,
-    .tool-bar-spacer {
+    .tool-bar-btn, .tool-bar-spacer {
```
ericcornelissen, line 65: Is this a compiling issue?

Personally, I prefer separate selectors to be on separate lines (which is a style consistently used in this file so far) as it clearly distinguishes descender selectors from separate selectors. I.e. the difference between these two:

```css
.tool-bar-btn .tool-bar-spacer { } /* style .tool-bar-spacer if it is a descendent of .tool-bar-button */
.tool-bar-btn, . tool-bar-spacer { } /* style .tool-bar-spacer and .tool-bar-button */
```

Why did you decide to change these (and not others, e.g. line 30 and 31)?

styles/tool-bar.less
```diff
@@ -38,6 +38,7 @@
     }
   }
 
+  // Set the margin between the buttons (that are not at the end)
```
ericcornelissen, line 41: Comment seems a little redundant to me, the LESS is pretty clear I would say :thinking: In fact, the LESS is removing the left/top margin and not setting any margins

styles/ui-variables.less
```diff
@@ -0,0 +1,85 @@
+// This file has fallback variables. It specifies the variables themes
```
ericcornelissen, line 1: I'm guessing this is what is causing less compiler problem? If so, I'm not against adding it, but it seems counter productive to me as these styles may not work with the user selected theme.

Do you have an example of other projects including this file?

styles/tool-bar.less
aminya: The standard has them in the same line:
https://www.w3schools.com/cssref/css_selectors.asp

I didn't go through all the lines. If consistency matters we can do that. Personally, I don't care.

styles/ui-variables.less
aminya: If this is the case, we can remove it. On Atom Dark without this file, the spacers were hidden, and I couldn't build the less file. 
No I don't know what others do. Maybe there is another way.

styles/tool-bar.less
aminya: You are right. My background isn't in web development, so it wasn't clear to me at first glance. Since the sentence was a little complicated, I thought it is worth mentioning the function. 

What comment do you suggest here instead?

styles/ui-variables.less
aminya: I added a screenshot the OP that shows the problem that is solved.

styles/tool-bar.less
ericcornelissen: > The standard has them in the same line:
https://www.w3schools.com/cssref/css_selectors.asp

First, standards are not style guides, and second a big reason why that standard (which it is not, it is more akin to documentation) has them on a single line is because it is in a table.

I had hoped the reasoning for my preference would make sense to you (i.e. being able to distinguish to very similar but vastly different CSS selectors), but if not I don't mind putting all selectors on a single line.


> I didn't go through all the lines. If consistency matters, we can do that for all the lines. Personally, I don't care.

I don't necessarily care about consistency. What I will say is that as a reviewer I do care that all changes in a Pull Request work towards the same goal. Because, again, is this a compiling issue?

I'm sorry if I'm being too nit-picky here :sweat_smile: Sure, for this small PR such small changes don't really matter, but I hope that you can see that in a larger PR these kinds of changes add up. I won't block this PR for these changes, but I hope that you see my point and  that you can learn from it in the future :smiley:

styles/tool-bar.less
ericcornelissen: In that case it's fine if the comment is there, changing "set" to "configure" would help but it's not necessary.

styles/ui-variables.less
ericcornelissen: > I added a screenshot the OP that shows the problem that is solved.

Thanks, that clarifies your motivation :smile: 

> If this is the case, we can remove it. On Atom Dark without this file, the spacers were hidden.

For your information, you can fix that using [your personal stylesheet](https://flight-manual.atom.io/using-atom/sections/basic-customization/#style-tweaks). I understand that you  might want to fix it for everyone using Atom Dark, but in that case I would create a fix in Atom Dark.


> and I couldn't build the less file.
>
> No, I don't know what others do. Maybe there is another way to link to Atom's default theme instead. Since less doesn't have a compiler config here, I don't know how it is possible. Maybe there is a CLI command for that.

It is not clear to me why you want to compile the less. As far as I know you don't need to do that as Atom will compile it for you at runtime, and it doesn't seem to run in to any problems doing that.

aminya: So does the Atom's compilation result change depending on the active theme?

aminya:
It seems that this PR isn't useful in a general case since it will override the active theme. If that's the case, we can close this.

PR #300, opened 2020-05-28
Remove activation hook

aminya:
Tool-bar late loading does not look good. It makes the Atom look bulky. I think 20ms is not that long to defer it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,8 +35,5 @@
         "1.0.0": "provideToolBar"
       }
     }
-  },
-  "activationHooks": [
-    "core:loaded-shell-environment"
-  ]
+  }
 }
--- a/spec/tool-bar-spec.js
+++ b/spec/tool-bar-spec.js
@@ -34,8 +34,9 @@ describe('Tool Bar package', () => {
       toolBarService = pack.mainModule.provideToolBar();
     });
 
-    atom.packages.triggerDeferredActivationHooks();
-    atom.packages.triggerActivationHook('core:loaded-shell-environment');
+    // if activation hook was being used
+    // atom.packages.triggerDeferredActivationHooks();
+    // atom.packages.triggerActivationHook('core:loaded-shell-environment');
   });
 
   describe('@activate', () => {


ericcornelissen:
Do you mean that the toolbar is being loaded in quite late and it suddenly just appears? I'm not entirely against removing it, but I'm personally not in favor of it. 20ms is not much but 1) that might be more on a slower computer and 2) if every package maintainer thinks that it adds up quickly.

Some alternatives we could consider:

- Remove the activation hook but update our implementation to get that 20ms down even further
- Add an animation for when the toolbar get's loaded, I suspect that will make it look less "bulky"

aminya:
Yes. It appears suddenly. I don't know how to add an animation. If you know, that can help. 

For performance, I can improve the loading time. That is on my todo list. 👍

aminya:
@suda Could you merge this? It makes Atom look much better.

suda:
Hi Amin, the activation hook [was added on purpose](https://github.com/suda/tool-bar/commit/92b4ec02a0e63da9ade56bb393efd2b5bf7cacd8) to improve Atom's performance.
I agree with adding the animation on showing the toolbar, and a CSS transition should be enough to create "slide-in" animation.

aminya:
Hi Wojtek! In #304 and #302, I improved the actual loading time of the package. The toolbar gets almost twice fast. I think this activation hook is not necessarily useful anymore. 

Note that I don't know how to add this CSS animation.

ericcornelissen:
> Hi Wojtek! In #304 and #302, I improved the actual loading time of the package. The toolbar gets almost twice fast (90ms to 45ms). I think this activation hook is not necessarily useful anymore.

In that case it would definitely make more sense to merge those PRs before merging this one :wink: If for whatever reason those PRs don't get merged we don't want this one having been merged already.


> Note that I don't know how to add this CSS animation.

That's fine, someone else can make that contribution if we decide not to merge this PR.

suda:
I agree we can add the animations later. And other PRs by Amin give a great performance boost anyway!

> The toolbar gets almost twice fast (90ms to 45ms)

Yes because it's trying to be loaded in parallel with Atom's core making it a bit slower instead. The goal of the hook was to give Atom all resources to bootstrap and then let it decide what to initialize in what order. We shouldn't try to squeeze ourselves into this process.

aminya:
@suda Can we merge this? Trying the trick in #306 was not successful for me.

suda:
As I said, there was a reason for adding the activation hook and I wouldn't like to remove it. We should come up with a nicer implementation of #306 instead of going against Atom's recommendation of using the hook IMO.

aminya:
@suda 

Activation hook is very useful for packages that don't need to load a feature right away. For example, they just work in the text editor, etc. But toolbar should appear when atom loads, and if does not, it makes atom look weird (compared to other IDEs)

I mentioned in the other repository too:

I think here we are doing things that are not common in other software/IDEs. All the usual software, load their toolbar normally during the loading time. We don't need to circumvent doing this!

 Tool-bar is, in my opinion, a crucial part of Atom, and it is so useful that I want to include in atom-community builds: https://github.com/atom-ide-community/atom/projects/4#card-41191288

suda:
You have a fair point. Lets merge this in then :+1:

aminya:
Thanks! :tada:

PR #301, opened 2020-05-28
Refactor to decrease code duplication and complexity

ericcornelissen:
This refactors the code base based on [the issues from Code Climate](https://codeclimate.com/github/suda/tool-bar/issues) regarding code duplication and code complexity. This should get us from a **C** to a (not perfect) **A** :smiley: 

When reviewing this Pull Request, I'd suggest going through the commits one-by-one to see if you agree with every refactoring step that I took. I think that looking at the final diff may be a bit confusing and the motivation for certain refactoring choices may be unclear.

### DIFF
--- a/lib/items/tool-bar-button-view.js
+++ b/lib/items/tool-bar-button-view.js
@@ -35,28 +35,11 @@ class ToolBarButtonView extends ToolBarItem {
       this.putAtEnd();
     }
 
-    if (options.icon) {
-      this.addIcon();
-    }
-
-    if (options.text) {
-      this.addText();
-    }
-
-    if (options.tooltip) {
-      this.subscriptions = new CompositeDisposable();
-      this.subscriptions.add(
-        this.addTooltip(this.options.tooltip, this.options.callback)
-      );
-    }
-
-    if (options.color) {
-      this.addColor();
-    }
-    if (options.background) {
-      this.addBackgroundColor();
-    }
-
+    this.addIcon();
+    this.addText();
+    this.addTooltip();
+    this.addColor();
+    this.addBackgroundColor();
     this.addClasses();
     this.addOnMouseDown();
     this.addOnClick();
@@ -79,6 +62,10 @@ class ToolBarButtonView extends ToolBarItem {
 
   /** Add an icon for the button using built-in icons. */
   addIcon () {
+    if (!this.options.icon) {
+      return;
+    }
+
     if (this.options.iconset) {
       if (this.options.iconset.startsWith('fa')) {
         this.classNames.push(this.options.iconset, `fa-${this.options.icon}`);
@@ -92,6 +79,10 @@ class ToolBarButtonView extends ToolBarItem {
 
   /** Adds a text/html to the button */
   addText () {
+    if (!this.options.text) {
+      return;
+    }
+
     if (this.options.html) {
       this.element.innerHTML = this.options.text;
     } else {
@@ -105,14 +96,18 @@ class ToolBarButtonView extends ToolBarItem {
    * @param {ButtonOptions.callback | null} callback
    * @returns {Disposable} a disposable tooltip
    */
-  addTooltip (tooltipOptions, callback = null) {
+  addTooltip () {
+    if (!this.options.tooltip) {
+      return;
+    }
+
     let tooltip;
-    if (typeof tooltipOptions === 'string') {
+    if (typeof this.options.tooltip === 'string') {
       tooltip = {
-        title: tooltipOptions
+        title: this.options.tooltip
       };
     } else {
-      tooltip = tooltipOptions;
+      tooltip = this.options.tooltip;
     }
 
     if (!tooltip.hasOwnProperty('placement')) {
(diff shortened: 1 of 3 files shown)


lib/items/tool-bar-button-view.js
```diff
@@ -79,6 +62,10 @@ class ToolBarButtonView extends ToolBarItem {
 
   /** Add an icon for the button using built-in icons. */
   addIcon () {
+    if (!this.options.icon) {
+      return;
+    }
+
```
aminya, line 68: I am not personally a fan of this type of refactoring. I deliberately wrote the code like this. I prefer checking the options outside of the function, and then only call the function if necessary.

lib/items/tool-bar-button-view.js
```diff
@@ -233,29 +218,28 @@ function getCallbackModifier (callback, {altKey, ctrlKey, shiftKey}) {
...
+    Left: 'right'
+  };
 
-  return toolbarPosition === 'Top' ? 'bottom'
-    : toolbarPosition === 'Right' ? 'left'
-      : toolbarPosition === 'Bottom' ? 'top'
-        : toolbarPosition === 'Left' ? 'right'
-          : null;
+  const toolbarPosition = atom.config.get('tool-bar.position');
+  return tooltipPlacement[toolbarPosition] || null;
```
aminya, line 242: This change is a better approach! Thanks.

lib/items/tool-bar-button-view.js
```diff
@@ -153,20 +146,12 @@ class ToolBarButtonView extends ToolBarItem {
...
     this.enabled = enabled;
   }
 
   setSelected (selected) {
-    if (selected) {
-      this.element.classList.add('selected');
-    } else {
-      this.element.classList.remove('selected');
-    }
+    this.element.classList.toggle('selected', selected);
```
aminya, line 154: Thanks for this. I didn't know about `toggle`.

lib/items/tool-bar-button-view.js
```diff
@@ -105,37 +96,39 @@ class ToolBarButtonView extends ToolBarItem {
...
-      typeof callback === 'string'
+      typeof this.options.callback === 'string'
     ) {
-      tooltip.keyBindingCommand = callback;
+      tooltip.keyBindingCommand = this.options.callback;
     }
 
-    return atom.tooltips.add(this.element, tooltip);
+    this.subscriptions = new CompositeDisposable();
+    this.subscriptions.add(atom.tooltips.add(this.element, tooltip));
```
aminya, line 124: Please revert this. We here create a subscription inside the tooltip, while the same subscription can be used by other parts of the code. I was seeing the future here.

lib/items/tool-bar-button-view.js
```diff
@@ -92,6 +79,10 @@ class ToolBarButtonView extends ToolBarItem {
 
   /** Adds a text/html to the button */
   addText () {
+    if (!this.options.text) {
+      return;
+    }
+
```
aminya, line 85: Same comment as https://github.com/suda/tool-bar/pull/301/files#r432136923

lib/items/tool-bar-button-view.js
```diff
@@ -105,37 +96,39 @@ class ToolBarButtonView extends ToolBarItem {
    * @param {ButtonOptions.callback | null} callback
    * @returns {Disposable} a disposable tooltip
    */
-  addTooltip (tooltipOptions, callback = null) {
+  addTooltip () {
+    if (!this.options.tooltip) {
+      return;
+    }
```
aminya, line 102: Same comment as https://github.com/suda/tool-bar/pull/301/files#r432136923

lib/items/tool-bar-button-view.js
```diff
@@ -105,37 +96,39 @@ class ToolBarButtonView extends ToolBarItem {
...
-  }
-
-  /** Add background color to the button */
-  addBackgroundColor () {
-    this.element.style.background = this.options.background;
+  /** Set a style on the button */
+  setStyle (style, value) {
+    if (value) {
+      this.element.style[style] = value;
+    }
```
aminya, line 131: Please take the if check out, and then this will be good.

lib/items/tool-bar-button-view.js
ericcornelissen: Sure, already found it a bit odd that the `CompositeDisposable` was previously created inside the if-branch for the tooltip, but I refactored it like this to stay consistent with the original implementation.

lib/items/tool-bar-button-view.js
ericcornelissen: Let me say up front that I'm okay with reverting this changes if you don't agree with the following :slightly_smiling_face: 

Firstly, it is not at all uncommon to check if the input is valid in other (including typed) programming languages, many programming languages have something similar to a `null` value that you nearly always need to take into account when writing a function.

Second, regarding the input/output type stability. This is not at all a problem in this case, as the function does not return anything. And even if it should return something, this pattern is still preferred over checking validity at the function call if it is possible to return a default value that always makes sense (sometimes called a [null object](https://en.wikipedia.org/wiki/Null_object_pattern)).

Lastly, the overhead of calling a few functions a couple of times (the number of buttons in the toolbar to be exact) is not something we as programmers working in a high-level language like JavaScript should worry about in comparison to making code readable and maintainable. first because the overhead is very low and second because the compiler or runtime environment can optimize it.

aminya: I agree with your points that this may not be applicable to this situation particularly and we don't see the issues I stated, but this is a general style that I always follow. If you want to write the checks compact you do something like this too:
```js
this.options.icon ?  this.addIcon() : ;
```

ericcornelissen: I'm even less of a  fan of writing it compactly in that way, as it obfuscates what you're trying to achieve (if you want to use an if-statement, use an if-statement and not a ternary operator).

If the original way is the way you always write it and that's how you want to do it here, that's completely fine by me. Let me know here and I will update the PR.

However, if you agree with my points I'm not sure why you would stick with our original style. But that's a discussion that should not block this PR from getting merged 🙂

aminya: In this case, this is just a minor styling change. I will let this to be up to you.

ericcornelissen: In that case let's merge it like this :smiley:

PR #302, opened 2020-05-29
Rollup - improve loading time

aminya:
- This makes the tool-bar load. reduces the actual loading time (if activation hook was not enabled)

**Note**: benchmark is done on top of https://github.com/suda/tool-bar/pull/304

After rollup:

Before (considering no drawGutter): 
image

For registering run the following:
```
npm install
npm run ppublish
apm publish patch
```


rollup.config.js
```diff
@@ -0,0 +1,48 @@
...
+
+    // so Rollup can convert externals to an ES module
+    commonjs(),
+];
+
+// minify only in production mode
+if (process.env.NODE_ENV === 'production') {
+    plugins.push(
+        // minify
+        terser({
```
ericcornelissen, line 18: how about importing this function with the name `minify` and removing the comment on the previous line?

.github/workflows/ci.yml
```diff
@@ -27,6 +27,7 @@ jobs:
       - name: Install APM dependencies
         run: |
           apm ci  # uses locked module. use `apm install` for non-locked
+          npm run build
```
ericcornelissen, line 30: Why not put this into a separate step?

```suggestion
      - name: Build
        run: npm run build
```

ppublish.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
```
ericcornelissen, line 2: Just curious, did you write this script yourself?

- If not: would it be possible to download it on-demand?
- If yes: what is your reasoning for checking if `.gitignore` exists and erroring if it does not?

package.json
```diff
@@ -8,22 +8,36 @@
...
   },
   "scripts": {
     "lint": "eslint . --fix --ext js,json && remark *.md",
-    "test": "atom --test spec"
+    "test": "atom --test spec",
+    "clean": "shx rm -rf dist",
+    "debug": "npm run clean && shx cp -r lib dist && env NODE_ENV=development env BABEL_ENV=development babel dist --out-dir dist",
+    "dev": "npm run clean && env NODE_ENV=development env BABEL_ENV=production rollup -c -w",
+    "build": "npm run clean && env NODE_ENV=production env BABEL_ENV=production rollup -c",
+    "ppublish": "bash ./ppublish.sh",
```
ericcornelissen, line 23: Does one need to manually execute this before publishing?

Not sure if that's the best solution. Have you considered using the [npm script](https://docs.npmjs.com/misc/scripts) `prepublishOnly` (which, for the record, I'm not 100% works with `apm`). Or maybe even `preinstall`, completely avoiding the need to include the `dist` folder in the project?

lib/touch-bar-manager.js
```diff
@@ -1,7 +1,8 @@
-const {TouchBar, nativeImage} = require('electron').remote;
+import {remote} from 'electron';
+const {TouchBar, nativeImage} = remote;
```
ericcornelissen, line 2: Not sure if the current JS version we're targetting supports this, but have you considered using [deconstructing nested objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Nested_object_and_array_destructuring)?

```suggestion
import {remote: {TouchBar, nativeImage}} from 'electron';
```

<sup>Hope I didn't mess up the syntax here</sup>

rollup.config.js
aminya: It is better to keep the plugin name.

ppublish.sh
aminya: I have written the script myself. If I upload it somewhere, how we download it then?

This script exists because we want to put `dist` in gitignore. If gitignore does not exist, this script is pointless.

package.json
aminya: Unfortunately, apm is not able to do that. This is the pattern I have used with all the other atom packages.

lib/touch-bar-manager.js
aminya: I tried this, but it does not work.

aminya:
> A few comments on the implementation below - do you have some numbers on how much this improves the loading time?


> 
> Also, is it the intention that the contents of `dist/` end up in the repo at some point? If so, could you add a `.gitattributes` file with the following contents:
> 
> ```
> * text=auto
> 
> # don't diff machine generated files
> lib/ -diff
> package-lock.json -diff
> ```

Yes, we have to include dist folder. That is the limitation of `apm`. 😞

aminya:
I found an interesting thing:
These two lines of the code take 30ms for their execution!
image

ppublish.sh
ericcornelissen: >  have written the script myself. If I upload it somewhere, how we download it then?

Then it's not necessary. If you're still interested we could upload it as a gist and use either something like curl, but preferably an NPM package for simplicity, to download it.

> This script exists because we want to put dist in gitignore. If gitignore does not exist, this script is pointless.

I would say it's fair to assume we have a `.gitignore` in this project, but I guess it doesn't hurt to check.

rollup.config.js
ericcornelissen: Why exactly?

`terser` is meaningless to me, `minify` is not. If I want to know what we're using to minify I can just look at where minify is imported and I see it's terser.

lib/touch-bar-manager.js
ericcornelissen: Did you check that is not due to my incorrect syntax? Since you're using Babel to compile this code I don't see why we couldn't use a JavaScript version that supports this, right?

ericcornelissen:
> But for that we need to include ui-variables. If you are OK with that we can do that.

As per #299, that seems like a bad idea to me.

> Yes, we have to include dist folder. That is the limitation of apm. :disappointed:

Ow well, okay :+1:

rollup.config.js
aminya: There are many plugins that do minification. It is common to keep the original name of the plugin in case people want to apply changes.

See https://github.com/TrySound/rollup-plugin-terser#usage

ppublish.sh
aminya: OK. I can make this an npm package. When I made one, I will create another PR to replace this.

lib/touch-bar-manager.js
aminya: What is the correct syntax?

aminya:
I added a script to rollup iconsets. That gives us another 10ms! 😃

lib/touch-bar-manager.js
ericcornelissen: For me the following works if I run it in my browser or Atom console it works just fine.

```js
let { remote: { TouchBar, nativeImage } } = { remote: { TouchBar: "foo", nativeImage: "bar" } };
console.log(TouchBar, nativeImage);
// ==> "foo bar"
```

aminya: I think the issue is deconstructing when importing. That is the error I was getting. If you can find a babel plugin that does that we can do it.

suda:
Nice! I agree that copying `ui-variables` is not ideal and with rolling up both code and iconsets, it might be enough.

lib/touch-bar-manager.js
ericcornelissen: Yes, it seems that this is not possible for import statements. Alas :disappointed:

rollup.config.js
ericcornelissen: I'm fine leaving it like this, I'm just suggesting something that I would argue is better in the long run (regardless of what is "common", also I don't think this would confuse anyone)

suda:
I'd also prefer the `dist` files to be in repo so all this dance of monkey patching the repo for `apm publish` could be removed in favor of simple `npm version` hook like this: https://github.com/particle-iot/particle-api-js/blob/master/package.json#L23

iconsets/rollup-iconsets.js
```diff
@@ -0,0 +1,7 @@
+// This file is the entry file for css rollup
```
ericcornelissen, line 1: This comment seems to conflict with the filename. The filename is telling me that "This file is the entry file for rolling up iconsets", not CSS in general.

Which was your intention and could you update the filename/this comment accordingly?

rollup.config.js
```diff
@@ -0,0 +1,66 @@
...
+            ecma: 2018,
+            warnings: true,
+            compress: {
+                drop_console: false,
```
ericcornelissen, line 24: Why do you want to maintain `console` statements in a production environment?

aminya:
@suda The issue with only using `git add` is that once someone runs `npm run build`, those files will be tracked by git and will show up in the changelogs and commits (although being in gitignore).

This bash script circumvents that issue, and only adds the dist files through `npm run ppublish`, and after that git will not track them.

iconsets/rollup-iconsets.js
aminya: Fixed in https://github.com/suda/tool-bar/pull/302/commits/a717e6a6854644b9ccb0a791bc6e181fe97bded6

rollup.config.js
aminya: Fixed in https://github.com/suda/tool-bar/pull/302/commits/a717e6a6854644b9ccb0a791bc6e181fe97bded6

aminya:
I made an npm package for ppublish script. https://github.com/aminya/build-commit
I will publish it tomorrow and using that instead of the bash script here.

aminya:
@suda Could you merge this?

Issue #303, opened 2020-05-30
[performance] Replace element.offsetHeight 


aminya:
I found an interesting thing: This line of code takes 30ms for its execution! Can't we do something about this?! 

![image](https://user-images.githubusercontent.com/16418197/83321560-227fc300-a216-11ea-9718-c3f3cb04e5e3.png)

This is only used for comparing it with `scrollHeight`. Can't we assume that this is true all the time?
https://github.com/suda/tool-bar/blob/master/lib/tool-bar-view.js#L185-L195

I disabled all the drawGutters and the loading time is greatly reduced! We should fix this line of code.

I changed the code to this and it works without any issues and the performance issue is gone!
```js
  drawGutter () {
    this.element.classList.remove('gutter-top', 'gutter-bottom');
        
    const scrollHeight = this.element.scrollHeight;

      if (this.element.scrollTop > 0) {
        this.element.classList.add('gutter-top');
      }
      if (this.element.scrollTop < scrollHeight) {
        this.element.classList.add('gutter-bottom');
      }
}
```

I see other people have the same issues, but there is no solution https://github.com/pelotoncycle/resize-observer/issues/5

ericcornelissen:
Would seem intuitively true, did you check that this does not reintroduce #156?

aminya:
@ericcornelissen Only in light theme and when positioned on the left, but that is because of `classList.add('gutter-bottom')`. 

Changing draw gutter to this solves both the performance and shadowing problem. I don't think we need to sacrifice the performance for something very minor.
```
  drawGutter () {
    this.element.classList.remove('gutter-top', 'gutter-bottom');

      if (this.element.scrollTop > 0) {
        this.element.classList.add('gutter-top');
      }
    }
```

suda:
The gutter shadow indicates that there are more icons in the tool-bar even if they are not visible. It will change if the window changed its size. I would suggest making it optional so users have the choice of turning it off :)

aminya:
Making it optional with default option being false seems good. 

However, isn't there any way to not use `element.offsetHeight` for the optional settings?

ericcornelissen:
I would suspect not, as its value is probably only available after the DOM has loaded. Perhaps we can look into executing this function in a non-blocking way (using Promises or workers) so that it doesn't affect the package load time significantly?

ericcornelissen:
> Making it optional with default option being false seems good.

[...]

PR #305, opened 2020-06-03
useGutter option

aminya:
Fixes https://github.com/suda/tool-bar/issues/303


suda:
Nice! That was quick 😃

aminya:
> Nice! That was quick 😃

Let's nominate this for the fastest time from issue to PR prize 🤣

PR #308, opened 2020-06-10
don't always set `tooltip.keyBindingCommand`

### LINKED ISSUE #307: [Performance] improve atom.tooltips.add
You would be surprised if you know that most of the time spent for creating the buttons, is spent in adding the tooltips to them (not the icons, etc)! We might want to write a faster version of the original 
https://github.com/suda/tool-bar/blob/937f75719a1002af5b207c073b6e45dd82617f44/lib/items/tool-bar-button-view.js#L121

For 30 buttons, this becomes ~6ms.
image

aminya:
- Fixes #307 
- Almost doubles the speed of button creation (X2 faster)
- See https://github.com/suda/tool-bar/issues/307#issuecomment-641719848 for the reason

Button creation time:

**After**: 
2020-06-10 00_01_00-Timecop — C__Users_yahyaaba_Documents_GitHub_JavaScript_tool-bar — Atom

**Before**:
image

This is not shown in the loading time of `tool-bar` since other packages add buttons using the service, but instead, it can be easily measured using `window.performance.now()`

### DIFF
--- a/lib/items/tool-bar-button-view.js
+++ b/lib/items/tool-bar-button-view.js
@@ -112,12 +112,6 @@ class ToolBarButtonView extends ToolBarItem {
       tooltip.placement = getTooltipPlacement();
     }
 
-    if (!tooltip.hasOwnProperty('keyBindingCommand') &&
-      typeof this.options.callback === 'string'
-    ) {
-      tooltip.keyBindingCommand = this.options.callback;
-    }
-
     this.subscriptions.add(atom.tooltips.add(this.element, tooltip));
   }
 


ericcornelissen:
I'm personally not a fan of this, as it actually changes the tool-tip as seen here:

| Before | After |
| --- | --- |
| before | after |

I would be fine with adding it as an option though.

aminya:
> I'm personally not a fan of this, as it actually changes the tool-tip as seen here:
> 
> Before	After
> before	after
> I would be fine with adding it as an option though.

We can certainly make a PR to Atom, and see why this key-binding adding takes time.

We can also add the key-binding to the actual text instead of going through Atom's code.
```js
tooltip.title = tooltip.title + this.options.callback
```

ericcornelissen:
> We can certainly make a PR to Atom, and see why this key-binding adding takes time. 

I prefer that 100% over just removing this feature from the tool-bar. But it's probably because it needs to find the keybind value in a rather large set of callback possibilities, though that doesn't mean it cannot be optimized.

> We can also add the key-binding to the actual text instead of going through Atom's code.
> > tooltip.title = tooltip.title + this.options.callback

That won't work. The callback value, in the case if-branch you removed is entered, is **not** the keybind string itself, it is the command name.

aminya:
> > We can certainly make a PR to Atom, and see why this key-binding adding takes time.
> 
> I prefer that 100% over just removing this feature from the tool-bar. But it's probably because it needs to find the keybind value in a rather large set of callback possibilities, though that doesn't mean it cannot be optimized.
> 
> > We can also add the key-binding to the actual text instead of going through Atom's code.
> > > tooltip.title = tooltip.title + this.options.callback
> 
> That won't work. The callback value, in the case if-branch you removed is entered, is **not** the keybind string itself, it is the command name.

I see. Thanks for the clarification. We can look into the actual function. The performance is not that bad though, but we may able to boost it a bit.

ericcornelissen:
> I see. Thanks for the clarification. We can look into the actual function. The performance is not that bad though, but we may able to boost it a bit.

That means we can close this PR, right?


Again, though, I'm not against adding an option that disables this feature 😉

aminya:
Let's keep it open for future reference. If I close it, it might get lost 😄

PR #310, opened 2020-07-18
Dynamic mdi icons

aminya:
This defers the loading of mdi icons until they are called. It improves the loading time (50ms->35ms).

image

image

### DIFF
--- a/lib/items/tool-bar-button-view.js
+++ b/lib/items/tool-bar-button-view.js
@@ -1,31 +1,6 @@
 import {CompositeDisposable} from 'atom';
 import { ToolBarItem } from './tool-bar-item';
-
-
-// Convert the CSS class name of mdi icons to the JS name
-function mdiIconName(str) {
-  // https://github.com/Templarian/MaterialDesign-JS/blob/master/build.js#L5
-  let name = str.replace(/(-\w)/g, (matches) =>  matches[1].toUpperCase());
-  name = `${name[0].toUpperCase()}${name.slice(1)}`;
-  return `mdi${name}`;
-}
-
-let mdi; // mdi iconSet
-async function createMDIIcon(iconName, iconSize) {
-  if (!mdi) {
-    mdi = await import('@mdi/js')
-  }
-  const iconData = mdi[mdiIconName(iconName)];
-  const icon = document.createElement("div");
-  icon.style.verticalAlign = "bottom"
-  icon.innerHTML = `
-    <svg style='width:${iconSize}px; height:${iconSize}px' viewBox='0 0 24 24'>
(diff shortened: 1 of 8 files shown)


aminya:
@ericcornelissen @suda I cannot seem to be able to propagate `button` style (size, colors, etc) to the icon. I would appreciate any help here.

lib/items/tool-bar-button-view.js
```diff
@@ -1,6 +1,32 @@
...
+  return "mdi" + str.split('-').map((p)=>capitalize(p)).join('')
+}
+
+let mdi; // mdi iconSet
+async function createMDIIcon(iconName) {
+  const svg = document.createElementNS("http://www.w3.org/2000/svg", "svg");
+  svg.setAttribute('width','auto');
+  svg.setAttribute('height','auto');
+  svg.classList.add('tool-bar-btn')
+  const path = document.createElementNS("http://www.w3.org/2000/svg", "path")
```
aminya, line 17: An alternative for this that I tried and works similarly:
```js
const icon = document.createElement("div");
icon.innerHTML = `
<svg>
<path d=${mdi[mdiIconName(this.options.icon)]} />
<svg/>
return icon
`

```

ericcornelissen:
> I cannot seem to be able to propagate button style (size, colors, etc) to the icon (svg). I would appreciate any help here.

Could you provide a working example outside of Atom (in [Codepen](https://codepen.io/) or a similar service) that doesn't style the SVG the way you expect it to, that would make it a little easier to understand what isn't working and how experimenting with how it can be resolved.

What I can say based just on your comment: Note that for the color you probably need to use the CSS attribute `fill` instead of `color` to add color the SVG and for the size you need to take into account the viewBox of the SVG.

aminya:
> What I can say based just on your comment: Note that for the color you probably need to use the CSS attribute `fill` instead of `color` to add color the SVG and for the size you need to take into account the viewBox of the SVG.

Yes I should do that instead. I thought SVG is like other HTML elements.

ericcornelissen:
> Yes, I should do that instead. I thought SVG is like other HTML elements. Isn't there a way to wrap SVG in an HTML element that allows CSS formatting?

As far as I know, the way you would want it to work is only possible with icon fonts.

aminya:
@ericcornelissen This now works. Now we need to fix the tests! I cannot query the svg element anyhow. 😞

aminya:
Besides this PR, I saw another very cool framework which is called iconify. I am trying to learn how it works. Using that we can expand the number of supported icons and also improve the speed.

 https://github.com/iconify/iconify/issues/31

Issue #237, opened 2018-08-14
MacBook Pro Touchbar virtual buttons blank


w-hc:
https://github.com/facebook/nuclide/issues/1582

This is an issue that appears in this downstream package bundle.

suda:
Hi! Thanks for filing this. I'll try to fix it in the next round of bugfixes!

PR #312, opened 2020-07-26
Update to current TouchBar Electron API

thatcomputerguy0101:
On my installation, the touch bar didn't originally work. After looking at the Electron documentation, I noticed that the TouchBar API is still experimental. Comparing the code of this library to a working example, I have determined that the fix is to wrap 'items' within an object when it is passed to the TouchBar constructor. This PR contains that change.

### DIFF
--- a/lib/touch-bar-manager.js
+++ b/lib/touch-bar-manager.js
@@ -76,7 +76,7 @@ export class TouchBarManager {
     const items = this.buttons
       .sort((a, b) => a.priority - b.priority)
       .map(button => button.touchButton);
-    let touchBar = new TouchBar(items);
+    let touchBar = new TouchBar({items});
     atom.getCurrentWindow().setTouchBar(touchBar);
     this._updateDebounce = null;
   }


thatcomputerguy0101:
Also, I may be wrong, but I believe this fixes the issue described in #237.

aminya:
I do not have a Mac to test this. If anyone of the other maintainers has they can review this.

ericcornelissen:
> Thanks for this. I do not have a Mac to test this. If anyone of the other maintainers has they can review this.

I don't so I can't test that either. @thatcomputerguy0101, is [this the documentation](https://www.electronjs.org/docs/api/touch-bar) on which the change is based? Did you try if the changes fixes the problem locally? If so, then I would be happy to merge this Pull Request.


> Also, I may be wrong, but I believe this fixes the issue described in #237.

Was your problem similar to what is described in that issue? If so I think it is safe to close that issue after this is merged 👍

thatcomputerguy0101:
> I don't so I can't test that either. @thatcomputerguy0101, is [this the documentation](https://www.electronjs.org/docs/api/touch-bar) on which the change is based? Did you try if the changes fixes the problem locally? If so, then I would be happy to merge this Pull Request.

Yes, the options parameter appears to have changed to an object. I have attached a screenshot of what the touch bar looks like with some buttons after I made this change locally.

Touch Bar Shot 2020-07-26 at 2 20 35 PM

> > Also, I may be wrong, but I believe this fixes the issue described in #237.
> 
> Was your problem similar to what is described in that issue? If so I think it is safe to close that issue after this is merged 👍

After looking at that issue in more detail, I think that the issue fixed here is different. Before I made the change, the only button on the touch bar was the default escape button, and it had nothing to do with the visibility state of the toolbar. Also, I never experienced the "ghost buttons" described in that issue. However, it could have been an older problem with Electron that got fixed since the TouchBar API is still under development.

ericcornelissen:
Alright, thanks for the followup @thatcomputerguy0101. In that case I will leave this for @suda to review, in case they can test this on a Mac with a touch bar.

jpcranford:
I'm not @suda, but I can confirm this fully fixes the issue, and by only adding { } around one little thing.

PR #326, opened 2020-10-21
fix deactivate

UziTech:
deactivate throw the following error on windows.

```
Error deactivating package 'tool-bar' TypeError: Cannot read property 'destroy' of null
```

### DIFF
--- a/lib/tool-bar.js
+++ b/lib/tool-bar.js
@@ -20,14 +20,10 @@ async function useTouchBar() {
 }
 
 export function deactivate() {
-  if (toolBarView) {
-    toolBarView.destroy();
-    toolBarView = null;
-  }
-  if (touchBarManager) {
-    touchBarManager.destroy();
-    touchBarManager = null;
-  }
+  toolBarView?.destroy();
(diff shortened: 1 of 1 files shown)


lib/tool-bar.js
```diff
@@ -20,10 +20,14 @@ async function useTouchBar() {
...
-  touchBarManager.destroy();
-  touchBarManager = null;
+  if (toolBarView) {
+    toolBarView.destroy();
+    toolBarView = null;
+  }
+  if (touchBarManager) {
+    touchBarManager.destroy();
+    touchBarManager = null;
+  }
```
aminya, line 30: Our babel config supports [optional chaining](https://github.com/suda/tool-bar/blob/b372ff7b296db6ca563fe5e781447d20ea30d9fd/babel.config.js#L9), so this nicer syntax would work:

```suggestion
  toolBarView?.destroy();
  toolBarView = null;
  touchBarManager?.destroy();
  touchBarManager = null;
```

aminya (approved):

PR #327, opened 2020-10-29
Change folder structure

idleberg:
I understand these changes are highly subjective, so please let me explain. The current folder structure is ambiguous about which files are meant for editing. So here is what I did:

- rename `lib/` to `src/` as this is a very common pattern
- rename `dist/` to `lib/` as this a common pattern with Atom packages (but with less impact)

It's a little change, but I expect it to make it easier for potential contributors to find their way around.

PS: No offence if you don't agree 😅


ericcornelissen:
I definitely agree with the lib to src renaming. The dist to lib is fine with me, but dist is clear as well in my opinion and not uncommon in my experience.

idleberg:
> The dist to lib is fine with me, but dist is clear as well in my opinion and not uncommon in my experience.

That's true, both are very clear and common. I guess what I was trying to refer to is that the official package generator uses `lib/`.

I will follow with whatever the majority agrees on and edit/undo those commits, if necessary.

aminya (changes requested):
Please keep dist as is. That's more clear and widely used by most of the tools.

package.json
```diff
@@ -16,11 +16,11 @@
   "scripts": {
     "lint": "eslint . --fix --ext js,json && remark *.md",
     "test": "atom --test spec",
-    "clean": "shx rm -rf dist",
-    "debug": "npm run clean && shx cp -r lib dist && env NODE_ENV=development env BABEL_ENV=development babel dist --out-dir dist",
+    "clean": "shx rm -rf lib",
+    "debug": "npm run clean && shx cp -r src lib && env NODE_ENV=development env BABEL_ENV=development babel lib --out-dir lib",
```
aminya, line 20:
```suggestion
    "clean": "shx rm -rf dist",
    "debug": "npm run clean && shx cp -r src dist && env NODE_ENV=development env BABEL_ENV=development babel lib --out-dir dist",
```

package.json
```diff
@@ -16,11 +16,11 @@
...
     "dev": "npm run clean && env NODE_ENV=development env BABEL_ENV=production rollup -c -w",
     "build": "npm run clean && env NODE_ENV=production env BABEL_ENV=production rollup -c",
-    "build-commit": "build-commit -o dist",
+    "build-commit": "build-commit -o lib",
```
aminya, line 23:
```suggestion
    "build-commit": "build-commit -o dist",
```

aminya (approved):

rollup.config.js
```diff
@@ -32,7 +32,7 @@ if (process.env.NODE_ENV === 'production') {
 export default [
     // Rollup lib
```
ericcornelissen, line 33: For consistency this should be updated to `// Rollup src`

```suggestion
    // Rollup src
```

